Issue #1, opened 2018-05-29
Data model


eliascodes:
Initial pass at the combined data model. To be iterated on.

<img width="1130" alt="screen shot 2018-05-29 at 16 21 04" src="https://user-images.githubusercontent.com/12719553/40668784-516be74c-635d-11e8-93da-d5d3645a47d3.png">

astroash:
Updated:

<img width="934" alt="screen shot 2018-05-29 at 5 57 54 pm" src="https://user-images.githubusercontent.com/14337958/40673548-fa0c3b48-6369-11e8-84c6-b0c387e140b1.png">

eliascodes:
Can probably remove the foreign key to `organisation` from `visit`, since it's linked via `visit_activity`.

astroash:
# v1.3
<img width="1242" alt="screen shot 2018-05-30 at 5 41 03 pm" src="https://user-images.githubusercontent.com/14337958/40734567-b59d7fae-6430-11e8-8be9-68c830a1c463.png">

eliascodes:
Also `name` seems to be a (semi-)reserved word, so we should be more specific with column names.

eliascodes:
Also remove `deleted` from `visit_activity`

eliascodes:
#v1.4

Fixed the stuff I mentioned, added the suggested permissions/roles structure from #2 and fixed `visit` table not linking to `user` table.

![screen shot 2018-05-31 at 14 30 53](https://user-images.githubusercontent.com/12719553/40785067-65d9c4aa-64df-11e8-98c2-0ec0d9a91dbb.png)

eliascodes:
I think so 🤔

astroash:
🔶 - all apps
🍏 - visitor app
🚙 - volunteer app
💜 - twine dashboard

eliascodes:
More:
* Subscriptions need an expiry
* Subscriptions needs to keep track of concept of "owner" and "beneficiary" of subscription since sector orgs buy and manage the subs on behalf of beneficiaries.
* Table to record trainings/workshops
* Optionally record post-code for visitors/volunteers
* Record meetings w/ local community members
* Funder table
* Funding programme table
* Record all CB logins

eliascodes:
Update:

![screen shot 2018-06-01 at 14 50 07](https://user-images.githubusercontent.com/12719553/40844208-3bf3139a-65ab-11e8-9904-00115b6445a3.png)

[...]

Issue #5, opened 2018-06-01
Deleting data for GDPR compliance


astroash:
After talking to Sonja, we will want to have functionality for soft deleting of all data. This will be a one way process that anonymises data by removing references to the users name, email and number.

This is not needed to organisation data.

PR #6, opened 2018-06-03
🗃 add initial DB migration and seeding scripts

eliascodes:
This is all based on our current (relatively good) understanding of the
data model, but is liable to change a little in future.

Related to #1


database/migrations/sql/20180603_create_initial_schema.sql
```diff
@@ -0,0 +1,461 @@
...
+
+  CONSTRAINT user_account_pk PRIMARY KEY (user_account_id)
+);
+CREATE TRIGGER update_user_account_modified_at BEFORE UPDATE ON user_account
+  FOR EACH ROW EXECUTE PROCEDURE update_modified_at_column();
+
+
+CREATE TABLE access_role (
+  access_role_id   SERIAL NOT NULL UNIQUE,
+  access_role_name VARCHAR NOT NULL UNIQUE,
```
astroash: Missing character limit, also we expecting the roles to change? Would this be better served with an enum?

database/migrations/sql/20180603_create_initial_schema.sql
```diff
@@ -0,0 +1,461 @@
...
+
+
+CREATE TABLE login_event (
+  login_event_id  SERIAL NOT NULL UNIQUE,
+  user_account_id INT NOT NULL,
+  created_at      TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+  deleted_at      TIMESTAMP WITH TIME ZONE,
+
+  CONSTRAINT login_event_pk         PRIMARY KEY (login_event_id),
+  CONSTRAINT login_event_to_user_fk FOREIGN KEY (user_account_id) REFERENCES user_account ON DELETE CASCADE
```
astroash: Do we want on delete cascade? All user data wants to be soft deleted #5, we could either remove the users row but keep their associated data, or keep their row but overwrite their personal information (name, email, number).

database/migrations/sql/20180603_create_initial_schema.sql
```diff
@@ -0,0 +1,461 @@
...
+  visit_feedback_id SERIAL NOT NULL UNIQUE,
+  organisation_id   INT NOT NULL,
+  score             INT NOT NULL,
+  created_at        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+  modified_at       TIMESTAMP WITH TIME ZONE,
+  deleted_at        TIMESTAMP WITH TIME ZONE,
+
+  CONSTRAINT visit_feedback_pk                 PRIMARY KEY (visit_feedback_id),
+  CONSTRAINT visit_feedback_to_organisation_fk FOREIGN KEY (organisation_id) REFERENCES organisation ON DELETE CASCADE,
+  CONSTRAINT visit_activity_valid_score        CHECK (score > -2 AND score < 2)
```
astroash: Do we want a 5 point feedback scale?

database/migrations/sql/20180603_create_initial_schema.sql
```diff
@@ -0,0 +1,461 @@
...
+  CONSTRAINT organisation_to_organisation_sector_fk FOREIGN KEY (organisation_sector_id) REFERENCES organisation_sector,
+  CONSTRAINT organisation_to_organisation_region_fk FOREIGN KEY (organisation_region_id) REFERENCES organisation_region,
+  CONSTRAINT organisation_to_funding_programme_fk   FOREIGN KEY (funding_programme_id)   REFERENCES funding_programme
+);
+CREATE TRIGGER update_organisation_modified_at BEFORE UPDATE ON organisation
+  FOR EACH ROW EXECUTE PROCEDURE update_modified_at_column();
+
+
+CREATE TABLE organisation_subordinate (
+  parent_id BIGINT NOT NULL,
```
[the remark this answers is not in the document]
eliascodes: Oversight

database/migrations/sql/20180603_create_initial_schema.sql
eliascodes: For fields that are not filled by user-supplied input, I think we can leave that character limit off. Means we wont constrain ourselves in future with arbitrary size limits that are a pain to change once in production.

database/migrations/sql/20180603_create_initial_schema.sql
eliascodes: Prefer the latter approach. If we don't cascade delete we'll either have unfulfilled constraints which will throw errors, or we'll have ex-foreign keys in tables which relate to a ghost row, which seems quite annoying if we ever want to pull that data together to use.

database/migrations/sql/20180603_create_initial_schema.sql
eliascodes: I don't know, do we?

database/migrations/sql/20180603_create_initial_schema.sql
```diff
@@ -0,0 +1,461 @@
...
+ * Extensions
+ */
+CREATE EXTENSION IF NOT EXISTS postgis;
+
+/*
+ * Enumerations
+ */
+CREATE TYPE genders             AS ENUM ('male', 'female', 'prefer not to say');
+CREATE TYPE turnover_bands      AS ENUM ('<100k', '100k-1m', '>1m');
+CREATE TYPE permission_type     AS ENUM ('own', 'own_organisation', 'own_beneficiaries');
```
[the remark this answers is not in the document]
eliascodes: Probs plural. The others are only plural so as not to clash with the column names in which they're used. Alternative naming suggestions welcome.

database/migrations/sql/20180603_create_initial_schema.sql
astroash: It's currently 3, we can check with Sonja.

eliascodes:
OK I think we're probably OK to get this re-reviewed and merged now because we're at the stage of building things on top of it (see #10)

database/migrations/sql/20180604_08_create_visits.sql
```diff
@@ -0,0 +1,74 @@
...
+  modified_at       TIMESTAMP WITH TIME ZONE,
+  deleted_at        TIMESTAMP WITH TIME ZONE,
+
+  CONSTRAINT visit_pk                   PRIMARY KEY (visit_id),
+  CONSTRAINT visit_to_user_fk           FOREIGN KEY (user_account_id)   REFERENCES user_account ON DELETE CASCADE,
+  CONSTRAINT visit_to_visit_activity_fk FOREIGN KEY (visit_activity_id) REFERENCES visit_activity
+);
+
+
+CREATE TABLE visit_feedback (
```
astroash: Online db schema calls this table `feedback`

database/migrations/sql/20180604_08_create_visits.sql
```diff
@@ -0,0 +1,74 @@
...
+  visit_feedback_id     SERIAL NOT NULL UNIQUE,
```
astroash: And this column `feedback_id`

database/migrations/sql/20180604_09_create_outreach.sql
```diff
@@ -0,0 +1,78 @@
...
+
+  CONSTRAINT outreach_campaign_pk                       PRIMARY KEY (outreach_campaign_id),
+  CONSTRAINT outreach_campaign_to_community_business_fk FOREIGN KEY (community_business_id) REFERENCES community_business ON DELETE CASCADE,
+  CONSTRAINT outreach_campaign_to_outreach_type_fk      FOREIGN KEY (outreach_type_id)      REFERENCES outreach_type
+);
+
+
+CREATE TABLE outreach_meeting (
+  outreach_meeting_id      SERIAL NOT NULL UNIQUE,
+  user_account_id          INT NOT NULL,
```
astroash: `user_id` on online db schema

database/migrations/sql/20180604_09_create_outreach.sql
```diff
@@ -0,0 +1,78 @@
...
+  modified_at           TIMESTAMP WITH TIME ZONE,
+  deleted_at            TIMESTAMP WITH TIME ZONE,
+
+  CONSTRAINT outreach_campaign_pk                       PRIMARY KEY (outreach_campaign_id),
+  CONSTRAINT outreach_campaign_to_community_business_fk FOREIGN KEY (community_business_id) REFERENCES community_business ON DELETE CASCADE,
+  CONSTRAINT outreach_campaign_to_outreach_type_fk      FOREIGN KEY (outreach_type_id)      REFERENCES outreach_type
+);
+
+
+CREATE TABLE outreach_meeting (
```
astroash: Online db schema has a column for `subject`

database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -0,0 +1,105 @@
...
+  entity        VARCHAR NOT NULL UNIQUE,
+  read_access   ENUM_permission_type NOT NULL,
+  write_access  ENUM_permission_type NOT NULL,
+  delete_access ENUM_permission_type NOT NULL,
+
+  CONSTRAINT permission_pk PRIMARY KEY (permission_id)
+);
+
+
+CREATE TABLE user_account_access_role (
```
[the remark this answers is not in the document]
eliascodes: This is a result of renaming some of these tables in this schema, because `user` and `role` are reserved words in SQL.

database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -0,0 +1,105 @@
...
+  deleted_at      TIMESTAMP WITH TIME ZONE,
+
+  CONSTRAINT user_account_access_role_to_user_fk         FOREIGN KEY (user_account_id) REFERENCES user_account ON DELETE CASCADE,
+  CONSTRAINT user_account_access_role_to_access_role_fk  FOREIGN KEY (access_role_id)  REFERENCES access_role  ON DELETE CASCADE,
+  CONSTRAINT user_account_access_role_to_organisation_fk FOREIGN KEY (organisation_id) REFERENCES organisation ON DELETE CASCADE,
+  CONSTRAINT user_account_access_role_unique_row         UNIQUE      (user_account_id, access_role_id, organisation_id)
+);
+
+
+CREATE TABLE access_role_permission (
```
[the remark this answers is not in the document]
eliascodes: As above

eliascodes:
@astroash I'm now thinking it would instead be better to only commit the migrations for the core (stable) tables. Those that are likely to change can be left out now and added back later since they are peripheral.

database/migrations/sql/20180604_04_create_organisations.sql
```diff
@@ -0,0 +1,108 @@
...
+
+  CONSTRAINT community_business_pk                              PRIMARY KEY (community_business_id),
+  CONSTRAINT community_business_to_organisation_fk              FOREIGN KEY (organisation_id)              REFERENCES organisation ON DELETE CASCADE,
+  CONSTRAINT community_business_to_community_business_sector_fk FOREIGN KEY (community_business_sector_id) REFERENCES community_business_sector,
+  CONSTRAINT community_business_to_community_business_region_fk FOREIGN KEY (community_business_region_id) REFERENCES community_business_region
+);
+
+
+CREATE TABLE organisation_invitation (
+  organisation_invitation_id SERIAL NOT NULL UNIQUE,
```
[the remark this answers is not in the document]
eliascodes: I'm removing this table from the initial migration, because we need clarity on TwinePlatform/hq#10 first.

database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -0,0 +1,105 @@
+/*
+ * User related tables
+ */
+
+CREATE TABLE user_account (
```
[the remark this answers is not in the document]
eliascodes: Don't think we need this any longer actually.

database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -0,0 +1,105 @@
...
+  deleted_at                       TIMESTAMP WITH TIME ZONE,
+
+  CONSTRAINT user_account_pk                  PRIMARY KEY (user_account_id),
+  CONSTRAINT user_account_sensible_birth_year CHECK       (birth_year > 1890 AND birth_year < date_part('year', CURRENT_DATE))
+);
+
+
+CREATE TABLE user_role (
+  user_role_id   SERIAL NOT NULL UNIQUE,
+  user_role_name VARCHAR NOT NULL UNIQUE,
```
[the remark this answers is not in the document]
eliascodes: This actually corresponds to the `access_role` table, which doesn't have timestamps. Will rename the table.

astroash:
☝️  I'm happy for this to be merged as is. I was just pointing out any differences so that it's documented if needed in future.

database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -0,0 +1,105 @@
...
+  CONSTRAINT user_secret_reset_pk                 PRIMARY KEY (user_secret_reset_id),
+  CONSTRAINT user_secret_reset_to_user_account_fk FOREIGN KEY (user_account_id) REFERENCES user_account ON DELETE CASCADE,
+  CONSTRAINT user_secret_reset_valid_expiry       CHECK       (expires_at > created_at),
+  CONSTRAINT user_secret_reset_valid_usage        CHECK       (used_at IS NULL OR (used_at < expires_at AND used_at > created_at))
+);
+
+
+CREATE TABLE login_event (
+  login_event_id  SERIAL NOT NULL UNIQUE,
+  user_account_id INT NOT NULL,
```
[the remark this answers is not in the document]
eliascodes: As above, this is a necessary renaming because of reserved words

database/migrations/sql/20180604_07_create_volunteer_tables.sql
```diff
@@ -0,0 +1,57 @@
...
+  activity_name         VARCHAR NOT NULL,
+  created_at            TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  modified_at           TIMESTAMP WITH TIME ZONE,
+  deleted_at            TIMESTAMP WITH TIME ZONE,
+
+  CONSTRAINT volunteer_activity_pk PRIMARY KEY (volunteer_activity_id)
+);
+
+
+CREATE TABLE volunteer_hours_log (
```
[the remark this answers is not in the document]
eliascodes: I think this name is better :)

eliascodes:
I'm updating the schema/diagram based on your comments. There is stuff I've thought of that I'm going to leave out of this PR anyway. Thanks for double checking everything 👍 

Will push shortly

eliascodes:
OK should be good to go @astroash

database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -6,27 +6,28 @@ CREATE TABLE user_account (
   user_account_id                  SERIAL NOT NULL UNIQUE,
   user_name                        VARCHAR(100) NOT NULL,
   user_secret                      VARCHAR(64) NOT NULL,
+  qr_code                          VARCHAR,
   gender                           ENUM_gender NOT NULL DEFAULT 'prefer not to say',
-  email                            VARCHAR(100),
+  email                            VARCHAR(100) NOT NULL UNIQUE,
```
astroash: If we're allowing users to sign up to multiple orgs and merge their accounts at a later point, this field doesn't want to be `UNIQUE`

https://github.com/TwinePlatform/hq/issues/11

eliascodes: Do'h

eliascodes: Actually, on second thoughts, maybe this can stay.

The Twine-Admin app enforces unique user emails, and in theory has a many-to-one relationship between users and organisations (although in practice I think this collapses to one-to-one).

What I meant to say in TwinePlatform/hq#11 is that we should enforce a 1-to-1 relationship in application code until we have time to support many-to-many.

All this means is an application-level check that the `user_account_id` column in the `user_account_access_role` table is unique.

If users want to sign up for another organisation, I don't see a problem with using a different e-mail. In fact that strikes me as quite natural.

The account merging process can be designed around this.

astroash: If users are using a different email address to sign up for another account, it will be difficult to merge their accounts automatically. This is the only piece of info that we collect that is unique.

eliascodes: > it will be difficult to merge their accounts automatically.

I don't think we need to do it automatically. And in fact, we may not _want_ to do it automatically. It should certainly be something the user opts in to, and they can do a little work to help us do the merge (e.g. enter all e-mail addresses they use, then we can send "are you sure" confirmation e-mails to all those addresses, merge all those that have been confirmed. Or we can design a number of alternatives).

astroash: Sold! 

Is this PR ready from your side?

eliascodes: I think so

PR #7, opened 2018-06-05
📝 Document New API

### LINKED ISSUE #2: Spiking authentication & authorisation
Research and adopt best practices, particularly for resource access control.

### LINKED ISSUE #4: Api Design
First attempt of api design:

Covers functionality for visitor app & native app (PHP app to be added). A number of previous named routes have been dropped in favor of HTTP methods, and `sort`/`search`/`page`/`[filter]` queries.

# New Api
## /sector
/sector
/sector/:id

/sector/:id/feedback
/sector/:id/visits
/sector/:id/activities

## /regions
/regions
/regions/:id

## /organisations
/organisations
/organisations/:id || whoami

/organisations/:id/feedback
/organisations/:id/visits
/organisations/:id/activities
/organisations/:id/users

/organisations/:id/volunteer_logs
/organisations/:id/meetings
/organisations/:id/outreach

## /users (does this want to be broken down for frontline?)
/users
/users/register
/users/login
/users/:id || whoami
/users/:id/password_reset

/users/:id/qr_code
/users/:id/qr_code/email
/users/:id/activities/:id (visit?)

/users/:id/volunteer_logs
/users/:id/volunteer_logs/:id
/users/:id/volunteer_logs/sync (potentially covered by POST)
/users/:id/meetings
/users/:id/meetings/sync (potentially covered by POST)
/users/:id/meetings/:id

astroash:
Fixes #4 
Fixes #2

**Changes**:
- new api design 
- document relationship between old api and new api
- document first draft of permission types


api.json
```diff
@@ -0,0 +1,761 @@
...
+          "scope": [],
+          "query": {
+            "fields": ["String"]
+          },
+          "response": {
+            "totalOrganisations": "Number",
+            "organisationsList":[{
+              "id": "Number",
+              "name": "String",
+              "logo_url": "String",
```
eliascodes: camelCase?

api.json
```diff
@@ -0,0 +1,761 @@
...
+        "GET": {
+          "description": "",
+          "auth": true,
+          "scope": [],
+          "query": {
+            "fields": ["String"]
+          },
+          "response": {
+            "totalOrganisations": "Number",
+            "organisationsList":[{
```
eliascodes: An idea to make the job of documenting requests/responses a little easier: define some types in the top-level of the JSON file (i.e. say what an "organisation" JSON-response looks like), and then simply reference that in the route definitions. For example:
```json
{
  "types": {
    "organisation": {
      "id": "Number",
      "name": "String",
    }
  },
  "routes": {
    "/regions": {
      "/:id/organisations": {
        "response": {
          "data": ["types.organisation"]
        },
      }
    }
  }
}
```

api.json
```diff
@@ -0,0 +1,761 @@
...
+              "sector": "String",
+              "registeredAt": "String",
+              "lastActiveAt": "String",
+              "modifiedAt": "String",
+              "turnoverBand": "String",
+              "360GivingId": "String",
+              "fundingProgramme": "String",
+              "fundingBody": "String",
+              "parentOrganisation": "String",
+              "childOrganisations": ["String"],
```
eliascodes: Including this kind of information requires additional queries. I thought about making it an opt-in field by default, but I guess that would complicate the backed logic a fair bit, and we control all the clients anyway so it's probably not that big of a deal.

TLDR: this is fine, just wanted to add this note.

api.json
```diff
@@ -0,0 +1,761 @@
...
+      "/": {
+        "GET": {
+          "description": "",
+          "auth": {},
+          "scope": {},
+          "query": {},
+          "response": {}
+        }
+      },
+      "/:id||me": {
```
eliascodes: I'm not sure this makes sense. `me` is really about the user, rather than the organisation or any other entity. 

Also remember that in the future, users (volunteers/visitors) may be related to several different organisations, so which would we choose?

To begin with we can have a constraint implemented in application logic that restricts 1 user to 1 organisation only (since it corresponds to assumptions made so far and application UI design so far), and then remove it down the line.

I think `/users/me/organisations` is better, and to start with it can just always return an array of one organisation.

What do you think?

api.json
```diff
@@ -0,0 +1,761 @@
...
+          "response": {
+            "feedbackScore": "Number",
+            "organisationId": "Number"
+
+          }
+        }
+      },
+      "/:id/users": {
+        "GET": {
+          "description": "Retreive all users in an organisation",
```
eliascodes: Should probably also have a filter by user type? Or instead have aliases `/:id/volunteers` and `/:id/visitors` and `/:id/admins`?

api.json
```diff
@@ -0,0 +1,761 @@
...
+              "volunteerLogId": "Number",
+              "userId": "Number",
+              "volunteerActivity": "String",
+              "duration": "Number",
+              "loggedAt": "String"
+            }]
+          }
+        }
+      },
+      "/:id/outreach": {
```
eliascodes: Is this gonna have more nested routes? Like `/:id/outreach/:oid`, `/:id/outreach/:oid/meetings`? Or is that handled some other way?

api.json
```diff
@@ -0,0 +1,761 @@
...
+          "scope": [],
+          "query": {
+            "fields": ["String"],
+            "filter": {
+              "outreachType": "String"
+            }
+          },
+          "response": {
+            "result?": [{
+              "outreachCampaignId": "Number",
```
eliascodes: Other responses use simply `id` as the key name of the top level entity. We should be consistent, and either use the full id key name everywhere or just use `id` everywhere.

api.json
```diff
@@ -0,0 +1,761 @@
...
+            "registeredAt": "String",
+            "lastActiveAt": "String",
+            "modifiedAt": "String"
+          }
+        },
+        "POST": {}
+      },
+      "/register": {
+        "POST": {
+          "auth": "Maybe - orgs/volunteers don't need auth but visitors do",
```
eliascodes: I think we can just reject registration attempts for orgs that don't have an active subscription? Would that cover it?

api.json
```diff
@@ -0,0 +1,761 @@
...
+          "description": "Retrieve details of a specific user",
+          "auth": true,
+          "scope": {},
+          "query": {
+            "fields": [],
+          },
+          "response": {
+            "id": "Number",
+            "name": "String",
+            "qrSecret": "String",
```
eliascodes: `qrSecret`? First, why are we sending secrets in the response? Second, the QR code is only relevant for visitors, so what does this mean for others?

api.json
```diff
@@ -0,0 +1,761 @@
...
+            "regionId": "Number",
+            "sectorId": "Number",
+            "organisationId": "Number",
+            "gender": "String",
+            "birthYear": "String",
+            "email": "String",
+            "phoneNumber": "String",
+            "emailConsent": "Boolean",
+            "smsConsent": "Boolean",
+            "qrCode": "String",
```
eliascodes: I think either this should be a conditional field based on whether the user is a visitor or not, or it should be a field that's only returned when we query explicitly for visitors, either through a "user type" filter or an alias like `/visitors/...`

api.json
```diff
@@ -0,0 +1,761 @@
...
+            "registeredAt": "String",
+            "lastActiveAt": "String"
+          }
+        },
+        "PUT": {
```
eliascodes: No delete method?

api.json
```diff
@@ -0,0 +1,761 @@
...
+          "body": {
+            "userId": "Number",
+            "email": "String",
+            "password": "String",
+            "confirmPassword": "String"
+          },
+          "response": {}
+        }
+      }
+      },
```
eliascodes: Is this an extra curly or a misaligned curly?

api.json
```diff
@@ -0,0 +1,761 @@
...
+     "/:id/visit_activities/:id": {
+      "POST": {
+        "description": "",
+        "auth": true,
+        "scope": [],
+        "query": {},
+        "response": {}
+      }
+      },
```
eliascodes: Ditto

api.json
```diff
@@ -0,0 +1,761 @@
...
+            "userId": "Number",
+            "email": "String",
+            "password": "String",
+            "confirmPassword": "String"
+          },
+          "response": {}
+        }
+      }
+      },
+     "/:id/visit_activities/:id": {
```
eliascodes: If this is nested under `/users`, not sure what the purpose of the route would be? Visit activities are owned by community businesses, not users.

api.json
```diff
@@ -0,0 +1,761 @@
...
+      "POST": {
+        "description": "",
+        "auth": true,
+        "scope": [],
+        "query": {},
+        "response": {}
+      }
+      },
+      "/:id/volunteers": {
```
eliascodes: I think naming should be clearer here. We're not retrieving volunteers, we're retrieving volunteering logs. `/:id/volunteer_logs` is better IMO.

api.json
```diff
@@ -0,0 +1,761 @@
...
+                "meetingType": "String",
+                "outreachPartner": "String",
+                "createdAt": "String",
+                "modifiedAt": "String"
+              }]
+            }]
+          }
+        }
+    },
+    "/email": {
```
eliascodes: plural

api.json
```diff
@@ -0,0 +1,761 @@
...
+        "POST": {
+          "description": "",
+          "auth": true,
+          "scope": [],
+          "query": {},
+          "response": {}
+        }
+      }
+    },
+    "/volunteers": {
```
eliascodes: Ah, this partially addresses one of my comments above 👍

api_migration.md
```diff
@@ -0,0 +1,121 @@
...
+`GET` replaces `/api/visitors/all` 👣 visitor app  
+`GET` replaces `/volunteers/organisation/{organisation}` ⏱ volunteer app  
+`GET` replaces `/volunteers/organisation360/{organisation}` ⏱ volunteer app  
+
+### /organisations/:id/visits
+`GET` replaces `/api/users/filtered`  👣 visitor app   
+`GET` replaces `/api/users/chart-all`  👣 visitor app  
+
+### /organisations/:id/volunteers 
+🤔 volunteers here reffers to volunteer logs. In two minds whether this is a better category over `/volunteer_logs`
```
eliascodes: I commented above that I think the naming should be clearer here and that your later suggestion is a better name.

However, we can also have a `/organisations/:id/volunteers` endpoint, but that endpoint should be concerned with users who are volunteers under the given organisation.

api_migration.md
```diff
@@ -0,0 +1,121 @@
...
+### /organisations/:id/visits
+`GET` replaces `/api/users/filtered`  👣 visitor app   
+`GET` replaces `/api/users/chart-all`  👣 visitor app  
+
+### /organisations/:id/volunteers 
+🤔 volunteers here reffers to volunteer logs. In two minds whether this is a better category over `/volunteer_logs`  
+`GET` replaces `/organisations/{organisation}/summary` ⏱ volunteer app  
+`GET` replaces `/logs/admin/{user}` ⏱ volunteer app  
+
+/organisations/:id/meetings (🤔 i don't think we need this)
```
eliascodes: Not sure, I guess drop it for now and we can stick it in later.

api_migration.md
```diff
@@ -0,0 +1,121 @@
...
+
+### /users/login
+`POST` replaces `/api/cb/login` 👣 visitor app  
+`POST` replaces `/api/admin/login` 👣 visitor app  
+`POST` replaces `/users/login`  ⏱ volunteer app  
+
+### /users/:id || me
+`GET` replaces `/api/user/details` 👣 visitor app  
+`GET` replaces `/api/admin/check` 👣 visitor app  
+`GET` replaces `/api/user/name-from-scan` 👣 visitor app   🤔 not sure if this should be covered here with qr params or under a `/users/:id/qr_code` route.
```
eliascodes: I think in my old design I was using a `/search` endpoint where you can search for entities by attributes. See https://github.com/TwinePlatform/twine-visitor/blob/master/docs/new_api.md#post-apivisitorssearch

api_migration.md
```diff
@@ -0,0 +1,121 @@
...
+
+### /users/:id/volunteers
+`GET` replaces `/logs/user/{user}/total` ⏱ volunteer app  
+`GET` replaces `/logs/user/{user}/total/days/{days}` ⏱ volunteer app  
+
+`GET` replaces `/logs/user/{user}` ⏱ volunteer app  
+`GET` replaces `/logs/user/{user}/total/date/{date}` ⏱ volunteer app  
+
+
+`POST` replaces `/logs/sync` ⏱ volunteer app 🤔 this could _potentially_ need its own route depending on how offline syncing is set up in the app
```
eliascodes: Agree, if the logic for determining what's already in the database is sufficiently complicated, we may just want to hive it all off into its own endpoint.

api_migration.md
```diff
@@ -0,0 +1,121 @@
...
+
+/users/:id/volunteers/:id  
+🤔 volunteers here reffers to volunteer logs. In two minds whether this is a better category over volunteers
```
eliascodes: As above

api_migration.md
```diff
@@ -0,0 +1,121 @@
...
+
+### /users/:id/outreach
+`GET` replaces `/meetings/user/{user}` ⏱ volunteer app  NB: this route is currently not being used
+
+`POST` replaces `/meetings/sync` ⏱ volunteer app 🤔 same as `/users/:id/volunteers/:id` sync question
```
eliascodes: As above

api_migration.md
```diff
@@ -0,0 +1,121 @@
...
+/users/:id/meetings
+
+/users/:id/meetings/sync (potentially covered by POST)
+/users/:id/meetings/:id
+
+/genders
+
+## /outreach
+### /outreach/type/:id
+`GET` replaces `/outreachChildTypes/parent/{id}` ⏱ volunteer app 🤔 return outreact types campaign targets
```
eliascodes: Dunno what this means.

api.json
astroash: Yeah, makes sense. Also `parentOrganisation` can probably be dropped now as it's the same as `fundingBody`

api.json
astroash: Outreach campaigns for organisations will be handled for in the org endpoint. This is just for outreach campaign target types.

.. although this is subject to change

api.json
astroash: This can be resolved by having `users/volunteers` and `users/visitors` endpoints

api.json
astroash: ☝️ same as above

api.json
astroash: 🔥

api.json
eliascodes: This isn't for target types, the response is for actual campaigns. There is a filter by type though.

api.json
eliascodes: Possibly, but we should certainly have the above anyway.

astroash:
@eliascodes this is ready for review. A number of questions and assumptions have been raised [here](https://github.com/TwinePlatform/twine-api/issues/2#issuecomment-396268406)

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+        }
+      },
+      "/password/reset": {
+        "POST": {
+          "description": "Reset password for a single user",
+          "auth": true,
+          "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin", "visitor", "volunteer", "volunteerAdmin"],
+          "scope": ["user_details-own:write"],
+          "body": {
+            "userId": "Number",
```
eliascodes: This doesn't work, how will the client know their user ID if they can't access the API (via their password)? Password resets always need a private token. Look at the definition of the existing password reset route in the visitor app.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+        },
+        "response": "types.adminUser"
+        },
+        "DELETE": {
+          "description": "Delete your own admin user account",
+          "auth": true,
+          "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin"],
+          "scope": ["user_details-own:delete"],
+          "body": {
+            "userName": "String"
```
eliascodes: Why require the user name? As a confirmation step? Maybe the password would make more sense? Or both.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          "body": {
+            "userId": "Number",
+            "email": "String",
+            "password": "String",
+            "confirmPassword": "String"
+          },
+          "response": null
+        }
+      },
+      "/admin/me": {
```
eliascodes: Why is this nested under `/users`? Could it not live under a top-level `/admins` and fulfill the same purpose?

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          "auth": true,
+          "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin"],
+          "scope": ["user_details-own:delete"],
+          "body": {
+            "userName": "String"
+          },
+          "response": null
+        }
+      },
+      "/volunteers/me": {
```
eliascodes: Same comment re: being nested under `/users`

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+        },
+        "response": "types.volunteerUser"
+        },
+        "DELETE": {
+          "description": "Delete your own volunteer user account",
+          "auth": true,
+          "intendedFor": ["volunteer"],
+          "scope": ["user_details-own:delete"],
+          "body": {
+            "userName": "String"
```
eliascodes: Same comment about confirmation step.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+        },
+        "response": "types.visitorUser"
+        },
+        "DELETE": {
+          "description": "Delete your own visitor user account",
+          "auth": true,
+          "intendedFor": ["visitor"],
+          "scope": ["user_details-own:delete"],
+          "body": {
+            "userName": "String"
```
eliascodes: Same comment about confirmation step

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          },
+          "response": null
+        }
+      },
+      "/:id/visit_activities/:id": {
+      "POST": {
```
eliascodes: 1. Indentation is not right
2. The RESTful convention is to make additions to collections using a `POST` to the endpoint describing the _entire_ collection. That means the appropriate way to create a visit should be `POST /visits` or `POST /organisations/:id/visits`. The current `/users/:id/visit_activities/:id` refers to a single entity, not a collection, so `POST`ing to it breaks convention.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+              "until": "String"
+            }
+          },
+          "response": {
+            "totalVolunteerHours": "Number",
+            "volunteerLogs": ["types.volunteerLog"]
+          }
+        }
+      },
+      "/:me/volunteer_logs": {
```
eliascodes: Replace `:me` with `me`.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+        "POST": {
+          "description": "Add volunteer log for your user account",
+          "auth": true,
+          "intendedFor": ["volunteer"],
+          "scope": ["volunteer_logs-parent:write"],
+          "query": {
```
eliascodes: We should be consistent in the use of `query` to describe query strings on get requests and `body` to refer to request bodies.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+            "organisationId":"Number",
+            "volunteerActvity": "String",
+            "duration": "Number",
+            "loggedAt": "String"
+          },
+          "response": "types.volunteerLog"
+        }
+      },
+      "/me/volunteer_logs/:id": {
```
eliascodes: No `GET`?

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          },
+          "response": "types.volunteerLog"
+        },
+        "DELETE": {
+          "description": "Delete volunteer log for your user account",
+          "auth": true,
+          "intendedFor": ["volunteer"],
+          "scope": ["volunteer_logs-parent:delete"],
+          "body": {
+            "organisationId":"Number",
```
eliascodes: Why are any of these necessary?

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+            "volunteerActvity": "String",
+            "duration": "Number",
+            "loggedAt": "String",
+            "delete": "Boolean"
+          },
+          "response": null
+        }
+      },
+      "/:id/outreach": {
```
eliascodes: Again, `outreach_campaigns` is a better way to describe this collection.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin", "volunteerAdmin"],
+          "scope": ["organisations_outreach-children:read", "organisations_outreach-own:read"],
+          "query": {
+            "fields": ["String"],
+            "filter": {
+              "outreachType": "String"
+            }
+          },
+          "response": {
+            "result": ["types.outreach"]
```
eliascodes: Again, probs best to split this as described (way) above into `outreach_campaings` and `meetings`

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+            "gender": "String"
+          },
+          "response": {
+            "genders": ["String"]
+          }
+        }
+      }
+    },
+    "/outreach": {
+      "/outreach/type/:id/campaing_targets": {
```
eliascodes: Typo `campaing`

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          "auth": true,
+          "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin", "volunteer", "visitor", "volunteerAdmin"],
+          "scope": ["dropdown_options:read"],
+          "query": {
+          },
+          "response": {
+            "genders": ["String"]
+          }
+        },
+        "POST": {
```
eliascodes: As with `sectors` and `regions`, probs only need `GET` for now.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+    },
+    "/volunteers": {
+      "description": "Alias for /users/** where user role is VOLUNTEER"
+    },
+    "/visitors": {
+      "description": "Alias for /users/** where user role is VISITOR"
+    },
+    "/genders": {
+      "/": {
+        "GET": {
```
eliascodes: As with `sectors` and `regions`, probs only need `GET` for now.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+      "postCode": "String",
+      "coordinates": "String",
+      "region": "String",
+      "sector": "String",
+      "registeredAt": "String",
+      "lastActiveAt": "String",
+      "modifiedAt": "String",
+      "turnoverBand": "String",
+      "360GivingId": "String",
+      "fundingProgramme": "String",
```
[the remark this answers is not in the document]
astroash: I've currently not refined optional queries in the api design.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+      "fundingBody": "String",
+      "subscription": "String"
+    },
+    "organisationTrainingSession": {
+      "trainingSessionAttendees": ["Number"],
```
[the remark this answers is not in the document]
astroash: Oops, this wasn't meant to be an array.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+      "email": "String",
+      "phoneNumber": "String",
+      "postCode": "String",
+      "emailConsent": "Boolean",
+      "smsConsent": "Boolean",
+      "registeredAt": "String",
+      "lastActiveAt": "String",
+      "modifiedAt": "String",
+      "organisation": {
+        "organisationId": "Number",
```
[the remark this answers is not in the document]
astroash: It felt more appropriate to add a reference to the org admins belong to rather than their full details. I don't think full org details need to be send as part of the user endpoint. This could either be removed or added to the other user types.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+    },
+    "visit": {
+      "id": "Number",
+      "activity": "String",
+      "visitDate": "String"
+    },
+    "outreach": {
+      "id": "Number",
+      "organisationName": "String",
+      "userOwner": "String",
```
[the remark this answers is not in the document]
astroash: Oops yeah, that should be on meeting instead!

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+      "id": "Number",
+      "meetingType": "String",
+      "outreachPartner": "String",
+      "createdAt": "String",
+      "modifiedAt": "String"
+    },
+    "volunteerLog": {
+      "id": "Number",
+      "volunteerActivity": "String",
+      "duration": "Number",
```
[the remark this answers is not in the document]
astroash: This is dependent on how the angular app is logging time.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          "auth": true,
+          "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin", "visitors", "volunteers", "volunteerAdmin"],
+          "scope": ["dropdown_options:read"],
+          "query": {},
+          "response": [{
+            "id": "String",
+            "sectorName": "String"
+          }]
+        },
+        "POST": {
```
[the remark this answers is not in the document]
astroash: 🤷🏽‍♀️ I can...

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          "intendedFor": ["twineAdmin", "fundingBody"],
+          "scope": ["organisations_details-children:read"],
+          "query": {
+            "fields": ["String"]
+          },
+          "response": {
+            "totalOrganisations": "Number",
+            "organisationsList": [{
+              "organisationDetails": "types.organisationDetails",
+              "childOrganisations": [{
```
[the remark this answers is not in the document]
astroash: 1. Many of these routes are designed to contain all the details for a relevant resource, with the client making use of the `fields` parameter for unwanted data. Whist creating the api it will become apparent if it's worth splitting this into its own route

2. Even though we no longer have a parent table in the database this relationship still exists, and is relevant to a number of places in the super admin dashboard for funding bodies such as subscriptions and data visualisation.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          "scope": ["dropdown_options:read"],
+          "query": {
+            "fields": ["String"]
+          },
+          "response": [{
+            "id": "String",
+            "regionName": "String"
+          }]
+        },
+        "POST": {
```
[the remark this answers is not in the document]
astroash: 😭

docs/api.json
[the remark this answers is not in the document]
eliascodes: OK, then we can just have `organisationId`. If we need the name we can make an additional request.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+            "region": "String",
+            "sector": "String",
+            "registeredAt": "String",
+            "lastActiveAt": "String",
+            "modifiedAt": "String",
+            "turnoverBand": "String",
+            "360GivingId": "String",
+            "fundingProgramme": "String",
+            "fundingBody": "String",
+            "parentOrganisation": "String",
```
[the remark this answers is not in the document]
astroash: It's not 🙃

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+      },
+      "/:id/visit_activities": {
+        "GET": {
+          "description": "Retrieve all visit activities for an organistion",
+          "auth": true,
+          "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin", "visitor"],
+          "scope": ["visit_activies-own:read", "visit_activies-children:read", "visit_activties-parent:read"],
+          "query": { 
+            "fields": "today",
+            "deleted": "Boolean"
```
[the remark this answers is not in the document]
astroash: This is optional, to include previously deleted activities.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+            "-1": "Number",
+            "0": "Number",
+            "1": "Number"
+          }
+        },
+        "POST": {
+          "description": "Send feedback for one organisation",
+          "auth": true,
+          "intendedFor": ["visitors"],
+          "scope": ["organisations_feedback-parent:write"],
```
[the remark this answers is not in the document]
astroash: Meant to be cbs not in _admin mode_ :godmode:

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+            "name": "String",
+            "gender": "String",
+            "birthYear": "String",
+            "email": "String",
+            "phoneNumber": "String",
+            "postCode": "String",
+            "emailConsent": "Boolean",
+            "smsConsent": "Boolean",
+            "secret": "String",
+            "invitationCode":"String"
```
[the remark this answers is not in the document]
astroash: I don't believe we have email confirmation codes. This is meant to be the admin code. I'll update the name.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+      "/:id": {
+        "GET": {
+          "description": "Retrieve details of a specific user",
+          "auth": true,
+          "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin"],
+          "scope": ["users_details-children:read"],
+          "query": {
+            "fields": ["String"]
+          },
+          "response": "types.visitorUser || types.volunteerUser || types.adminUser"
```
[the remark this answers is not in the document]
astroash: This is an endpoint to get a specific user. A generic user object can be returned with null fields for those that aren't relevant. We may also want to add a role key.

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+        },
+        "PUT": {
+        "description": "Update details of a specific user",
+        "intendedFor": ["twineAdmin", "fundingBody", "cbAdmin"],
+        "scope": ["users_details-children:write"],
+        "body": {
+          "name": "String",
+          "regionId": "Number",
```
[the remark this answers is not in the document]
astroash: Happy to leave these in until we make that choice.

docs/api.json
astroash: 🙃

docs/api.json
```diff
@@ -0,0 +1,1070 @@
...
+          },
+          "response": {
+            "totalOrganisations": "Number",
+            "organisationsList": [{
+              "organisationDetails": "types.organisationDetails",
+              "childOrganisations": [{
+                "id": "Number",
+                "name": "String"
+              }],
+              "trainingSessionsList": "types.organisationTrainingDetails"
```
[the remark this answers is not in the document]
astroash: 🔥

docs/api.json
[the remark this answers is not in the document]
eliascodes: We will have e-mail confirmation codes, see the account creation flow issue in hq.

astroash:
@eliascodes ready for review 🤞🏽

PR #28, opened 2018-06-20
✅ Test api routes using api.json

astroash:
~~**NB: This will not pass travis tests until TwinePlatform/eslint-config-twine#3 is merged**~~

**Changes:**
- update ts setting to include es7 lib
- add missing descriptions/auth to api.json
- add isImplemented key to api.json
- update linting to tslint via shared config

Relates #27

PR #30, opened 2018-06-22
📝 Update readme contents

astroash:
**Changes:**
- add missing md links in README contents


astroash:
@eliascodes, we're currently not installing specific versions of eslint-config-twine, so this will fail until #28 is merged

Issue #32, opened 2018-06-26
Survey data and push endpoint for Frontline


eliascodes:
After our meeting yesterday, it was clear that the two tables in the [Twine-Admin](https://github.com/TwinePlatform/Twine-Admin) app that are storing survey data are storing aggregated data that would be useful for visualising cross-organisation statistics. This needs to stay in the data model.

This data was being pushed to the admin app DB by FrontlineSMS on a real-time basis, so we need to accommodate this by adding a push endpoint to the API spec and generating an API key for FrontlineSMS to use when pushing. See the original controller for this [here](https://github.com/TwinePlatform/Twine-Admin/blob/master/app/Http/Controllers/API/SurveyAnswerController.php)

eliascodes:
Via e-mail from Frontline:

> **Twine survey median answer data**
> Every time an answer arrives through SMS into any Community Business workspace we, in addition to the usual survey processing, push the answer text to http://dashboard.twine-together.com/api/v1/survey-answers using a single (shared across all workspace) Reason-provided API Key.  The JSON looks like this:
> ```
> {
>    "apiKey": "${userData['api-key']}",
>        "payload": {
>            "answer_content": "<SMS TEXT>",
>            "answerUUID": “<ANSWER-UUID>",
>            "360GivingId": "${globalUserData['360GivingId']}"
>        }
> }
> ```
> I believe that Reason used to simply receive these and store them in the db; we were not yet set up to use this data.  So I think the first task is to simply connect up this pipe again so that the data is being saved.
>
> The next step will then be for the PHP server to have an API endpoint set up that can provide median statistics for each of the answers on request, or by pushing the data out regularly (TBC).  What is needed here is for the API to receive API Key (for auth) + UUID (which question) + 360GivingID (provides data such as the CB’s region which might be needed) and have a set of statistics calculated from the database data and returned; this may just be average or more (TBC).
>
> Here at Frontline we will then connect this up and present the median data on the same axes as the CB’s live survey answer data.
>
> **Central CLS data store**
> This will work in the same way as proposed above with the PHP server -> FrontlineSMS API bridge.  The data here is static and updated once a year.  The data will probably simply be numeric and associated again with each question’s UUID.
>
> FrontlineSMS would again send the same Key + UUID + 360ID so that regional data can be pulled and present that on the charts.

eliascodes:
After speaking with @SonjaTwine both these items seem to be of different priority.

[...]

PR #33, opened 2018-06-26
Data model updates

eliascodes:
Related to #32 

Various updates to the data model including a new migration for the survey related data pushed in from Frontline.

Update data model docs.

Add DB reset script.


database/index.js
```diff
@@ -1,6 +1,11 @@
 const fs = require('fs');
 const path = require('path');
-const { compose, last, head } = require('ramda');
+const knex = require('knex');
+const { compose, last, head, tap } = require('ramda');
+const { getConfig } = require('../config');
+
+// lazyPromiseSeries :: [PromiseLike (a)] -> Promise ([a])
```
astroash: what's `PromiseLike`?

eliascodes: Something that is compatible with the Promise/A+ interface but isn't a native `Promise` (e.g. the `knex.QueryBuilder`)

eliascodes:
> Also.. are we going to convert these js files to ts at some point?

Yes, I'll get to it after the data-sync stuff, or you can have a bash at it if you like.

PR #36, opened 2018-06-26
📝 Update config docs with correct postmark keys

astroash:


eliascodes:
Not merging just yet this so you can include other missing things from the setup guide.

astroash:
@eliascodes I think this is good to go.

PR #42, opened 2018-07-03
Data model updates

eliascodes:
Some minor updates based on work on the data sync tool.

Some inconsistencies between the schema diagram and the migrations have also been fixed.


database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -12,13 +12,15 @@ CREATE TABLE gender (
...
+  user_password                    VARCHAR(64),
+  qr_code                          VARCHAR UNIQUE,
   gender_id                        INT NOT NULL,
-  email                            VARCHAR(100) NOT NULL UNIQUE,
+  email                            VARCHAR(100) UNIQUE,
   phone_number                     VARCHAR(20),
   post_code                        VARCHAR(10),
   birth_year                       INT,
+  is_email_confirmed               BOOLEAN NOT NULL DEFAULT false,
+  is_phone_number_confirmed        BOOLEAN NOT NULL DEFAULT false,
```
astroash: Are we going to be sending text message confirmations?

database/seeds/development/11_access_role.seed.js
```diff
@@ -0,0 +1,10 @@
+exports.seed = (knex) =>
```
astroash: Lol, I have this in my pr. I'm gunna have to rewrite all my test.

eliascodes: Should just need to do find/replace, no?

database/migrations/sql/20180604_06_create_users.sql
eliascodes: Maybe. Either way, we should know whether we know this, if you get me.

astroash: Nope, but it's not overly important right now.

PR #31, opened 2018-06-22
✨ Add email client interface

astroash:
~~_NB:_ TwinePlatform/eslint-config-twine#6 to be merged first~~

**Changes:**
- create email interface
- add templates as json
- add ramda helper functions
- add tests
- add postmark types as local file to make a few edits

Relates: #29

_Disclaimers:_
- assuming postmark is meant to throw errors in testing environments, cannot find much in their docs
- there's loads of ~~typescript errors~~ `any` type definitions

PR #45, opened 2018-07-03
Permissions module

### LINKED ISSUE #40: 🗃 Permission types
https://github.com/TwinePlatform/twine-api/blob/b1eb23f6e44f9f7d72bbb03095bb270c2c6c9886/database/migrations/sql/20180604_06_create_users.sql#L42-L46

`permission_type` is used for accessing resources dependent on a roles relationship to it, eg a cb admin accessing a visitors data with `parent` permission type. Currently not all resources have a permission type as this relationship is irrelevant eg `constants`. 

To work with our data model we can either:
- remove `NOT NULL` constraint from `permission_type`
- add a generic `permission_type` which is unrelated to the relationship between roles & resources eg `all`

### LINKED ISSUE #41: 🐛 migrate:teardown script not dropping all tables
When running `npm run migrate:teardown` in testing env,  frontline_account table is not dropped.

Running `npm run migrate:teardown`:
screen shot 2018-06-29 at 11 55 51 am

Running `npm run migrate:teardown`:
screen shot 2018-06-29 at 11 56 02 am
screen shot 2018-06-29 at 11 56 10 am

### LINKED ISSUE #46: Test command runs all tests twice
Discovered while reviewing #31 

The test command transpiles the `ts` files to `js`, then jest picks up the newly created `js` files and runs them too.

astroash:
#31 to be merged first

**Changes**:
- fix migrate:teardown script
- remove db logs in testing env
- create permissions section permissions module corresponding type definitions
- add test data to test db
- add tests
- add test env scripts
- rename permission_type -> permission_level

Fixes #41 
Fixes #40
Fixes #46


src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+  client
+    .select('access_role_id')
+    .table('access_role')
+    .where({ ['access_role_name']: role })
+    .then((res) => res[0].access_role_id);
+
+const permissionsInitialiser: PermissionsInitialiser = (config) => {
+  const client = knex(config);
+  return {
+    grantExisting: async ({ resource, permissionLevel, access, role }) => {
```
eliascodes: This is just so I can confirm my understanding of what this does: Grants requested access level on resource to a role, if that combination of (resource x access x permission level) already exists.

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+      const roleId = await getRoleId({ client, role });
+
+      const permissionIds = await getPermissionIds({
+        client,
+        resource,
+        permissionLevel,
+        access });
+
+      if (permissionIds.length === 1) {
```
eliascodes: I think this line suggests we're missing something in our DB constraints. Shouldn't the combination of (resource, access, permission level) have a unique constraint? Would there be any case where we'd need to have two rows with the same 3-tuple?

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+    .where({ ['access_role_name']: role })
+    .then((res) => res[0].access_role_id);
+
+const permissionsInitialiser: PermissionsInitialiser = (config) => {
+  const client = knex(config);
+  return {
+    grantExisting: async ({ resource, permissionLevel, access, role }) => {
+      const roleId = await getRoleId({ client, role });
+
+      const permissionIds = await getPermissionIds({
```
eliascodes: I think it's fine for this to throw instead of us performing checks on its output?

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+
+const getRoleId: GetRoleId = async({ client, role }) =>
+  client
+    .select('access_role_id')
+    .table('access_role')
+    .where({ ['access_role_name']: role })
+    .then((res) => res[0].access_role_id);
+
+const permissionsInitialiser: PermissionsInitialiser = (config) => {
+  const client = knex(config);
```
eliascodes: I think it's better to take the actual `knex` client as an argument here, because we'll want to share the same connection pool across the entire app.

It's different from the e-mail service, because the postmark client (AFAIK) doesn't create any remote connections when it's instantiated. Everything is done via HTTP requests to the postmark server, and only on method calls.

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+                ['access_role_id']: roleId,
+                ['permission_id']: permissionId,
+              })
+                .whereNotExists(client('access_role_permission').where({
+                  ['access_role_id']: roleId,
+                  ['permission_id']: permissionId,
+                }))
+            );
+        if (insertPermission.length === 0) {
+          throw new Error('Permission entry is already associated to this role');
```
eliascodes: Might be worth sub-classing our own custom errors (list `PermissionError` or something more specific).

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+        const createPermissionId = await client('permission')
+            .returning('permission_id')
+            .insert({
+              ['permission_entity']: resource,
+              ['permission_level']: permissionLevel,
+              ['access_type']: access,
+            });
+
+        const permissionId = createPermissionId[0];
+        const insertPermission = await client('access_role_permission')
```
eliascodes: Maybe this and the above can be combined into a single query, using a common table expression (CTE), using [with](https://knexjs.org/#Builder-with):
```js
const insertPermission = client('permission').returning('*').insert({...})
client
  .with('new_permission', client.raw(insertPermission.sql, insertPermission.bindings))
  .insert({
    ['access_role_id']: roleId,
    ['permission_id']: select('permission_id').from('new_permission')
  })
```
Basically I'm trying to get this SQL:
```sql
WITH (
  INSERT INTO permission (...) VALUES (...) RETURNING permission_id
) AS new_permission_id
INSERT INTO access_role_permission (...) VALUES (..., new_permission_id) RETURNING *
```
The benefit is that the above is run as a single, all-or-nothing query without having to resort to a full-on transaction. As the code is currently written, there's a small chance we could be left with an inconsistent state in the DB if something goes wrong halfway through.

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+            .returning('*')
+            .insert({
+              ['access_role_id']: roleId,
+              ['permission_id']: permissionId,
+            });
+        client.destroy();
```
eliascodes: As we switch to receiving a `knex` client as an argument, we should get rid of these calls to `destroy`, as they should be handled by the caller(s).

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+        resource,
+        permissionLevel,
+        access });
+
+      if (permissionIds.length === 1) {
+        const permissionId = permissionIds[0].permission_id;
+
+        const deleteRow = await client('access_role_permission')
+            .where({
+              ['access_role_id']: roleId,
```
eliascodes: Throughout all these methods, since `getRoleId` and `getPermissionIds` just return `knex.QueryBuilder` instances, you don't need to `await` them. You can just combine them all into a single query that uses sub-queries:
```js
client('access_role_permission')
  .del()
  .where({ 
    access_role_id: getRoleId({ client, role }),
    permission_id: getPermissionIds({ ... }),
  });
```
This way, `knex` should also error when either sub-query returns more than one row, (possibly) saving you from doing checks on outputs.

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+              ['permission_id']: permissionId,
+            })
+            .del();
+
+        if (deleteRow === 1) return deleteRow;
+        throw new Error('Permission entry is not linked to role');
+      }
```
eliascodes: I can't see whether the brackets line up here, but if `permissionIds.length !== 1`, it looks like we just do nothing and return `undefined`. Is that right?

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+            'access_role.access_role_id',
+            'access_role_permission.access_role_id')
+          .where({
+            ['access_role.access_role_name']: role,
+            ['permission.access_type']: access,
+            ['permission.permission_entity']: resource,
+            ['permission.permission_level']: permissionLevel,
+          });
+      const roleHasPermission = client.raw('SELECT EXISTS ?', [inner]);
+      const result = await roleHasPermission;
```
eliascodes: Why not combine this line with the one above?

src/auth/permissions/index.ts
```diff
@@ -0,0 +1,173 @@
...
+        'user_account_access_role.access_role_id',
+        'access_role.access_role_id')
+        .where({
+          ['user_account_access_role.user_account_id']: userId,
+          ['permission.access_type']: access,
+          ['permission.permission_entity']: resource,
+          ['permission.permission_level']: permissionLevel,
+        });
+      const userHasPermission = client.raw('SELECT EXISTS ?', [inner]);
+      const result = await userHasPermission;
```
eliascodes: As above

src/auth/permissions/permissions.d.ts
```diff
@@ -0,0 +1,117 @@
...
+  port: number,
+  database: string,
+  user: string,
+  password?: string,
+  ssl: boolean,
+}
+
+type DevelopmentConnection = string
+
+type PermissionsConfig = {
```
eliascodes: This should be provided by the `knex` typings, no?

src/auth/permissions/index.ts
eliascodes: If any operation consists of multiple DB queries, we should always think about data consistency, and whether we can combine the queries to be run as a single atomic block.

That usually means either sub-queries, CTEs or transactions.

package.json
```diff
@@ -21,15 +21,20 @@
   "homepage": "https://github.com/TwinePlatform/twine-api#readme",
   "main": "src/index.js",
   "scripts": {
-    "test": "jest",
+    "test": "NODE_ENV=testing jest",
+    "test:watch": "NODE_ENV=testing jest --watch",
     "lint": "tslint --project tsconfig.json",
     "lint:fix": "tslint --project tsconfig.json --fix",
     "migrate:latest": "knex migrate:latest",
+    "migrate:latest:test": "NODE_ENV=testing knex migrate:latest",
```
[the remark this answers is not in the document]
astroash: This was shorter to type as I was using them lots. I'm not fussed either way.

src/auth/permissions/index.ts
astroash: Yes

package.json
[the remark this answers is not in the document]
eliascodes: 🤷🏼‍♂️ lets leave them in then

src/auth/permissions/index.ts
astroash: This is checking that the permission (resource x permissionLevel x access) exists. I'm missing an else statement that throws an error if the combination does not exist.

eliascodes: Oh, OK. In which case the sub-query style that I mention in one of my later comments means this is probably unnecessary, since `knex` will throw if a sub-query returns `N !== 1` rows.

src/auth/permissions/index.ts
astroash: What do you mean? I'm checking how many rows are returned to see if the correct method has been used

src/auth/permissions/index.ts
astroash: I'll look into what the error messages from knex are like under, although I was going for module specific error messages which tell you why is went wrong and what you should use instead

src/auth/permissions/index.ts
astroash: By overwriting Error.name or prefixing the message with PermissionError?

eliascodes:
```js
class PermissionError extends Error {}
```

eliascodes: Not enough for it be worth the stuff in that SO answer. I've never had to do that before though. Anyway, ignore this, it was just a thought.

src/auth/permissions/index.ts
astroash: From what I can see, sql doesnt error if subqueries/where statements return nothing in delete statements. This means I can't return appropriate errors if this is done in one query.

database/index.js
```diff
@@ -57,35 +57,53 @@ exports.migrate = {
...
+    
+    const tables = await teardownClient('pg_catalog.pg_tables').select('tablename').where({ schemaname: 'public' });
 
     const queries = tables
       .map((t) => t.tablename)
       .filter((t) => t !== 'spatial_ref_sys')
-      .map(tap((t) => console.log(`Dropping table ${t}`)))
-      .map((tablename) => client.raw(`DROP TABLE IF EXISTS "${tablename}" CASCADE`))
+      .map(tap((t) => { 
+        if(config.env !== 'testing') console.log(`Dropping table ${t}`)
```
eliascodes: The permitted environment values are enumerated as constants in the config module, so lets import them and use them here instead of hardcoding the value.

database/index.js
```diff
@@ -57,35 +57,53 @@ exports.migrate = {
...
-        'ENUM_permission_type',
+        'ENUM_permission_level',
         'ENUM_access_type',
         'ENUM_invitation_status',
         'ENUM_subscription_status',
       ]
-        .map(tap((t) => console.log(`Dropping type ${t}`)))
-        .map((e) => client.raw(`DROP TYPE IF EXISTS ${e} CASCADE`))
+        .map(tap((t) => {
+          if(config.env !== 'testing') console.log(`Dropping type ${t}`)
```
eliascodes: As above

src/auth/permissions/__tests__/index.test.ts
```diff
@@ -0,0 +1,187 @@
...
+import * as knexInit from 'knex';
+
+describe('Permisions Module', () => {
+  const config = getConfig(process.env.NODE_ENV);
+  const knex = knexInit(config.knex);
+  const permissionsInterface = permissionsInitialiser(knex);
+  beforeAll(async() => {
+    await migrate.teardown({ client: knex });
+    await knex.migrate.latest();
+    await knex.seed.run();
```
eliascodes: Minor point, but I think we can remove this, since it'll be run in `beforeEach` anyway

src/auth/permissions/__tests__/index.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          role:'VOLUNTEER' });
+        expect(query.exists).toBe(false);
+      } catch (error) {
+        expect(error).toBeFalsy();
+      }
+    });
+    test('SUCCESS - returns true for matching permissions & user', async() => {
+      try {
+        const query = await permissionsInterface.roleHas({
+          resource: 'constants',
```
eliascodes: I think all these literal string values that are being passing as arguments should actually be constants that are imported from the permissions module. That's how it is in the initial design. It decouples meaning from value.

src/auth/permissions/permissions.d.ts
```diff
@@ -0,0 +1,100 @@
+import * as Knex from 'knex';
+
+type Dictionary<T> = { [k: string]: T }
+
+type Role =
```
eliascodes: If these were enums we could export them and use them in calling code instead of hardcoding the values.

src/auth/permissions/permissions.d.ts
```diff
@@ -0,0 +1,100 @@
...
+  | 'visit_logs'
+
+type PermissionLevel =
+  'own'
+  | 'child'
+  | 'parent'
+  | 'any'
+
+
+type ProductionConnection = {
```
eliascodes: This is unused, so is the one below it.

Issue #48, opened 2018-07-05
Two additional attributes/tags for users


SonjaTwine:
Hi,
I'm putting this here so you have the earliest warning possible - PTC is still figuring out the details.

We will need two additional attributes for **individual users** - this will be collected via the sign up page for the **visitor** app and the **volunteer** app.

Disability: Yes/No/Prefer not to say/NA (this will be an optional question)
Ethnicity: (Likely to be a longer list of varchar. I'm trying to get a shorter version of the 2001 Census categories.)

We can make the front end changes to the two sign up pages later. (i.e. adding two additional questions with dropdowns, plus info text related to GDPR)

The visitor app will then feature two additional graphs in "visits" that are pie charts showing the shares of each category for both of the above attributes.
(and Frontline, too, for volunteers)

However, we will not make front end changes for this on the individual user profile that CB admins can see on the visitor or volunteer app. That way, CB admins will only work with anonymised data.

Thoughts/concerns/questions?

eliascodes:
Thanks for the heads up. Sounds OK. The only thing I would say is that we should be aware that the more data we wish to collect from visitors _at signup_, the more burdensome/off-putting that process becomes for them.

If the number of fields increases much more in future it may be worth thinking about how to make the signup process at least _feel_ simpler/quicker, even if it isn't.

SonjaTwine:
Thanks!
Yup, agreed!  I'll add that to the bits I'm discussing with Jem in the first week of August.
Will update here when I have details on the contents of the ethnicity bit, and also time frames!

SonjaTwine:
Nope, apologies - following up!
What are the dependencies of this?

eliascodes:
Nothing really, we've just made some up for now.

SonjaTwine:
Great.
May I suggest adding "muggle", "wizard", "elf", "hobbit" if you haven't already...
No sorry, I'll get you that list.

astroash:
@edwardwptc do we have a list of ethnicities to be used on the twine platform?

edwardwptc:
The default choice would be the 2011 census list because if anyone has a problem with it, we can say it's the census choices and not ours... That list might be a bit long tho, Sonja said about shortening it but unless you have a reason to I'd just use it as is for the above reason. Ignore where it says "please specify", if you think that's ok?  

White
English/Welsh/Scottish/Northern Irish/British
Irish
Gypsy or White Traveller
Any other white background (please specify)

[...]

PR #49, opened 2018-07-06
Data model updates

eliascodes:
Related to #48 

Various data model updates, some based on #48, others necessary in the course of working on TwinePlatform/data-sync#15


database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -9,12 +9,30 @@ CREATE TABLE gender (
...
+);
+
+
 CREATE TABLE user_account (
   user_account_id                  SERIAL NOT NULL UNIQUE,
   user_name                        VARCHAR(100) NOT NULL,
   user_password                    VARCHAR(64),
   qr_code                          VARCHAR UNIQUE,
   gender_id                        INT NOT NULL,
+  disability_id                    INT NOT NULL,
```
astroash: Is this having a default?

database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -9,12 +9,30 @@ CREATE TABLE gender (
...
+  ethnicity_id                     INT NOT NULL,
```
astroash: Same as above

database/seeds/development/13_ethnicity.seed.js
```diff
@@ -0,0 +1,17 @@
+exports.seed = (knex) =>
+  knex('ethnicity').del()
+    .then(() =>
+      knex('ethnicity').insert([
+        { ethnicity_name: '🎶' },
```
astroash: 👌

database/seeds/testing/13_ethnicity.seed.js
```diff
@@ -0,0 +1,17 @@
+exports.seed = (knex) =>
```
astroash: This file is going to conflict with my pr due to file numbering. Not bothered which goes first.

database/migrations/sql/20180604_06_create_users.sql
eliascodes: I think no, because it's an ID, and I don't want to rely on particular IDs corresponding to particular values. We'll enforce the default in application code.

database/seeds/testing/13_ethnicity.seed.js
eliascodes: I think you'll need to rename your seed files because your user records will need valid foreign key references.

database/migrations/sql/20180604_06_create_users.sql
astroash: Out of curiosity, why are we using id's rather than enums?

PR #57, opened 2018-07-11
Roles module

### LINKED ISSUE #34: Create permissions module
Create module to manage permissions and roles for users. Initial design can be found here:

https://github.com/TwinePlatform/twine-api/blob/master/docs/permissions.md#permissions-module

astroash:
**Changes:**
- Create roles modules
- Setup two testing scripts so one can run parallel and the other in series for db tests

Fixes #34


src/auth/roles/index.ts
```diff
@@ -0,0 +1,126 @@
+import { RolesInitialiser } from './types';
+const { lazyPromiseSeries } = require('../../../database');
+import * as knex from 'knex';
+
+const rolesInitialiser: RolesInitialiser = (client) => {
+  return  {
+    add: async({ role, userId, orgId }) => {
```
eliascodes: Think I prefer the syntax:
```ts
return {
  async add ({ role, userId, orgId }) => {},
  async remove () => {},
}
```

Also I'd prefer not to abbreviate `organisationId` to `orgId`

src/auth/roles/index.ts
```diff
@@ -0,0 +1,126 @@
...
+        const query = await client.insert({
+          user_account_id: userId,
+          organisation_id: orgId,
+          access_role_id: client.select('access_role_id')
+            .table('access_role')
+            .where({ ['access_role_name']: role }),
+        })
+        .into('user_account_access_role')
+        .returning('*');
+        return await  query[0];
```
eliascodes: The `await` here is unnecessary

src/auth/roles/index.ts
```diff
@@ -0,0 +1,126 @@
...
+        organisation_id: orgId,
+        access_role_id: client.select('access_role_id')
+          .table('access_role')
+          .where({ ['access_role_name']: to }),
+      })
+      .into('user_account_access_role');
+      try {
+        return await client.transaction((trx) => {
+          lazyPromiseSeries([deleteRow, addRow].map((q) => q.transacting(trx)))
+          .then(trx.commit)
```
eliascodes: Let's indent chained method calls

src/auth/roles/index.ts
```diff
@@ -0,0 +1,126 @@
...
+              'User is already associated with this \'from\' role at this organistion');
+
+          default:
+            throw error;
+        }
+      }
+    },
+    removeUserFromAll: async({ userId, orgId }) => {
+      const deleteRow = await client('user_account_access_role')
+        .where({ user_account_id: userId,
```
eliascodes: Can we format this object consistently?

src/auth/roles/types.ts
```diff
@@ -0,0 +1,17 @@
+import * as Knex from 'knex';
+import { Role } from '../permissions/types';
+import { Dictionary } from 'ramda';
+
+export type QueryResponse = Dictionary<any>;
+export type QueryResponseArray = Dictionary<any>[];
```
eliascodes: This is unnecessary I think

src/auth/roles/index.ts
```diff
@@ -0,0 +1,126 @@
+import { RolesInitialiser } from './types';
+const { lazyPromiseSeries } = require('../../../database');
```
eliascodes: If this function is useful in more than one place now, it's probs worth sticking it in a utils file or something somewhere.

src/auth/roles/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      if (!checkFromExists.rows[0].exists) {
+        throw new Error(`From role is not associated associated to this user`);
+      }
+
+      const deleteRow = client('user_account_access_role')
+        .where({ user_account_id: userId,
+          organisation_id: orgId,
+          access_role_id: client.select('access_role_id')
+          .table('access_role')
```
eliascodes: Indent

src/auth/roles/__tests__/index.test.series.ts
```diff
@@ -0,0 +1,151 @@
...
+          `Foreign key does not exist: Key (user_account_id)=(3) is not present in table \"user_account\".`
+        );
+      }
+    });
+  });
+
+  describe('::remove', () => {
+    test('SUCCESS - remove existing link between user and role', async() => {
+      try {
+        const result:any = await rolesInterface.remove(
```
eliascodes: Let's get rid of this

src/auth/roles/__tests__/index.test.series.ts
```diff
@@ -0,0 +1,151 @@
...
+  });
+
+  afterAll(async() => {
+    await knex.destroy();
+  });
+
+  describe('::add', () => {
+    test('SUCCESS - adds role for a user', async () => {
+      try {
+        const result:any = await rolesInterface.add({ role: Role.VOLUNTEER, userId:1, orgId: 1 });
```
eliascodes: And this

package.json
```diff
@@ -35,7 +37,7 @@
     "seed:run:test": "NODE_ENV=testing knex seed:run",
     "db:reset": "npm run migrate:teardown && npm run migrate:latest && npm run seed:run",
     "db:reset:test": "npm run migrate:teardown:test && npm run migrate:latest:test && npm run seed:run:test",
-    "cover": "jest --coverage",
+    "cover": "jest --runInBand --coverage",
```
eliascodes: I don't think we want to run everything in series. Can we separate out commands for those tests that run in series (basically anything touching the DB) and those in parallel (unit tests and others with limited dependencies)

astroash: We need to run all of our tests together to get an accurate coverage report. As the database tests need to be run in series, this is needed here.

Please add an issue to JEST to ask them why there is no configurable option for specific tests to be run in series.

src/auth/roles/index.ts
astroash: Shorthand syntax for method definitions is only for functions, rather than arrow functions. I'm not sure if there's any added benefit of functions over arrow functions here. Stylistically I find it slightly more readable using arrow functions as the method name is first.

eliascodes:
```js
return {
  async add = ({ x }) => {
    // ...
  }
}
```

astroash: Isn't this public class field syntax, which is currently polyfilled by babel (in classes)?

PR #58, opened 2018-07-11
Deploy to heroku

### LINKED ISSUE #54: Deploy
Worth getting it up now as opposed to at the end.

eliascodes:
Fixes #54 

### Changes
Nothing major that's not obvious from the diff or commit messages. 

One thing to note, though: because the database scripts are in JS and are not transpiled/moved into the build folder by the build script, they cannot refer to the typescript-version of the config module, and must instead point to the built JS version of it (in the build folder). This is fine as long as the build command has been run before hand. 

It's a temporary fix and can be removed once the database scripts are in TS too (low priority).


astroash:
@eliascodes 👇

PR #75, opened 2018-07-17
✨ Add external auth scheme

### LINKED ISSUE #60: Authentication strategies
Based on #32 it's clear that the API will need to support two distinct authentication strategies:
* Standard token-based authentication for regular use-cases
* Pre-authenticated tokens for use with trusted third parties (i.e. Frontline)

astroash:
**Changes**:
- create db table to hold api tokens
- add tests seed data
- create auth scheme for external users
- setup strategy with holding route
- add tests

fixes #60


src/api/v1/organisations/index.ts
```diff
@@ -17,4 +17,20 @@ export default [
       return '';
     },
   },
+  {
+    method: 'GET',
+    path: '/cls',
```
eliascodes: My thinking was that the `api/v1` directory would be organised into directories based on the root level path parameter, so all `/organisations/...` routes would go into and `organisations` directory.

So I think I would either put this in a `cls` directory or think about whether CLS data should be accessible under some (yet to be designed) survey endpoint (such as `/surveys/cls` or `/surveys/answers/cls` or similar).

src/auth/scheme/validate_external.ts
```diff
@@ -0,0 +1,44 @@
...
+
+type TokenResponse = { api_token: string, api_token_access: string };
+
+
+const findMatching = async (xs: TokenResponse [], token: string): ExternalStrategyResponse => {
+  if (xs.length < 1) return { isValid: false, credentials: {} };
+
+  const { api_token: apiToken = null, api_token_access: scope = null } = xs[0];
+
+  if (!apiToken || !scope) throw new Error('Error collecting api token data from database');
```
eliascodes: Both these fields are marked `NOT NULL` in the database so it doesn't seem necessary to make this check here.

src/auth/scheme/validate_external.ts
```diff
@@ -0,0 +1,44 @@
...
+  try {
+    const tokenResponses = await knex('api_token').select(['api_token', 'api_token_access']);
+    if (tokenResponses.length === 0) {
+      throw new Error('No api token data in database');
+    }
+
+    return findMatching(tokenResponses, token);
+  } catch (error) {
+    console.log(error);
+    return Boom.badImplementation('Error with route authentication for 3rd party clients');
```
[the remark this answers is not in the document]
astroash: Yes, I am only throwing an error is there is a problem with the database, if there are no matching tokens I return `isValid: false` which is picked up by the auth scheme and returns a `401`

src/auth/scheme/validate_user.ts
```diff
@@ -51,7 +51,7 @@ const validateUser: ValidateUser = async (decoded, request) => {
     return { isValid: false };
   } catch (error) {
     console.log(error);
-    return Boom.badImplementation('Error with server authentication');
+    return Boom.badImplementation('Error with route authentication for users');
```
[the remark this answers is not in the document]
astroash: This strategy should only throw if there's a problem with knex, bad `jwt`s are handled by the auth scheme, and empty `jwt`s return `isValid: false` on line 49.

src/api/v1/organisations/index.ts
astroash: Assuming this is achieved by making each group of endpoints a plugin and using [`options.routes.prefix`](https://hapijs.com/api#server.register())?

eliascodes: I was planning on only using the prefix option for the top-level API v1 plugin in order to prefix either `/api/v1` or `/v1` (depending on what URL we host it at) onto all the API routes.

I thought it would be good (for clarity and searchability) to require all route definitions to specify their full path.

src/auth/scheme/validate_external.ts
```diff
@@ -0,0 +1,41 @@
...
+    isValid: true,
+    credentials: { scope }, }
+  : findMatching(xs.slice(1), token);
+};
+
+const validateExternal: ValidateExternal = async ({ knex }, token, h) => {
+  try {
+    const tokenResponses = await knex('api_token').select(['api_token', 'api_token_access']);
+    if (tokenResponses.length === 0) {
+      Boom.unauthorized('No stored api token data');
```
eliascodes: Is there a `return` missing here?

Issue #76, opened 2018-07-18
Create database model objects


eliascodes:
We'll use these model objects as an abstraction layer on top of `knex` to present a simple API that hides away DB implementation details (e.g. the fact that user attributes like gender and disability are stored in different tables from the main user record).

* [x] User model
   * [x] Visitor model
   * [x] CB Admin model
   * [x] Volunteer model
   * [ ] Super Admin model
   * [ ] Twine Admin model
* [x] Organisation model
* [x] Community Business model
* [ ] Funding Body model
* [ ] Subscription model
* [ ] Survey model
* [ ] Outreach Campaign model
* [x] Visit model
* [x] Visit Activity model
* [x] Volunteer log model
* [ ] Volunteer Activity model
* [ ] Single-use token model

eliascodes:
Downgraded priority since the initial minimal set of models is done.

astroash:
@eliascodes the platform has widely differed from the initial vision. I am in favour of this issue being closed and cb models being created as and and when they are needed.

PR #77, opened 2018-07-18
User model

eliascodes:
Related to #76

OK so this is just _part_ of the user model. Have a look and tell me if you'd prefer if I add the rest of the user model in this PR, or if we use a different PR for that.

### Additions
* Beginnings of the user model
* Dead simple version of the first user API endpoint `GET /users`

### Changes
* Target es2017 when compiling, Node 10+ supports it
* Update dev script
* Fix badge in README


src/models/user.ts
```diff
@@ -0,0 +1,101 @@
...
+  'gender.gender_name': 'gender',
+  'ethnicity.ethnicity_name': 'ethnicity',
+  'disability.disability_name': 'disability',
+};
+const ModelToColumn = invertObj(ColumnToModel);
+
+/*
+ * Implementation of the UserCollection type
+ */
+export const Users: UserCollection = {
```
astroash: We're both using very different naming conventions. I think we need to pick a naming convention for this project as matter of priority. It's going to be a lot of unnecessary work doing this later down the line, and will complicate code throughout the app until this point.

src/models/util.ts
```diff
@@ -0,0 +1,35 @@
+import * as Knex from 'knex';
+import { intersection, compose } from 'ramda';
+import { CommonQueryOptions } from './models';
+
+
+export const Utils: {
+  [k: string]: (...a: any[]) => (q: Knex.QueryBuilder) => Knex.QueryBuilder
+} = {
+  limit: (n: number) => (q: Knex.QueryBuilder) => q.limit(n),
```
astroash: What does this type signature mean?

src/models/util.ts
```diff
@@ -0,0 +1,35 @@
...
+  offset: (n: number) => (q: Knex.QueryBuilder) => q.offset(n),
+  order: (c: string, d: string) => (q: Knex.QueryBuilder) => q.orderBy(c, d),
+};
+
+export const applyQueryModifiers = (p: Knex.QueryBuilder, opts: CommonQueryOptions) => {
+  const modifiers = intersection(Object.keys(opts), Object.keys(Utils));
+
+  const query = modifiers
+    .reduce((acc, k: keyof CommonQueryOptions) => {
```
astroash: Out of curiosity, why are you grabbing the keys of `Utils` and `CommonQueryOptions` separately for the same data?

eliascodes: `opts` may include zero or more of the keys in `CommonQueryOptions`, and `Utils` does not support `fields` (because it's dealt with in the `select`), so taking the intersection of the two means I will only iterate over keys actually in `opts` and actually supported by `Utils`.

I think that's what you were asking but am not sure.

src/models/util.ts
eliascodes:
```
limit :: Number -> QueryBuilder -> QueryBuilder
```
It just returns a modified `QueryBuilder` instance.

eliascodes:
Apologies, this PR has swollen by ~200 lines mostly because of trying to improve coverage.

src/models/__tests__/user.test.integration.ts
```diff
@@ -0,0 +1,33 @@
...
+
+describe('User Model', () => {
+  const config = getConfig(process.env.NODE_ENV);
+  const knex = Knex(config.knex);
+
+  afterAll(async () => {
+    await knex.destroy();
+  });
+
+  test('get :: no arguments gets all users', async () => {
```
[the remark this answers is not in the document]
eliascodes: Yeah, I think I want to add some seed data in the next PR once I get a better idea of whats needed for the tests on the write methods (update, add, delete).

src/models/user.ts
eliascodes: Are you talking about `Users` or `UserCollection`? The term "collection" comes from ORM-like libraries (e.g. Bookshelf) that use the term to describe objects that represent the entire set of a particular type of entity. I think I could have alternatively used "repository" which is used for the same thing in Laravel (PHP).

If you mean there's an inconsistency between this usage and the usage of "interface" in the roles/permissions objects, I suppose it depends on how whether you consider those objects to be model objects (i.e. something that is supposed to represent an entity in a way that hides how that entity is actually implemented in the database), or not.

The API they have at the moment is not consistent with the "Collection" interface defined in `models.d.ts`, and I'm not sure they need to be.

Maybe we can talk about this on hangouts, I don't think it can be easily resolved here.

tests/server.ts
```diff
@@ -0,0 +1,34 @@
+/*
+ * Test utility to create a slightweight server object
```
astroash: slight -> light

PR #79, opened 2018-07-19
Permissions seeds

astroash:
**Change**: _see comments_


database/seeds/development/11_access_role.seed.js
```diff
@@ -4,7 +4,9 @@ exports.seed = (knex) =>
       { access_role_name: 'VISITOR' },
       { access_role_name: 'VOLUNTEER' },
       { access_role_name: 'VOLUNTEER_ADMIN' },
+      { access_role_name: 'ORG_NON_ADMIN' },
```
eliascodes: What is this new user type for?

database/seeds/development/11_access_role.seed.js
```diff
@@ -4,7 +4,9 @@ exports.seed = (knex) =>
...
       { access_role_name: 'ORG_ADMIN' },
+      { access_role_name: 'FUNDING_BODY' },
```
eliascodes: Is this necessary because working out whether the organisation someone is an admin of is a funding body or a CB? In which case `ORG_ADMIN` now has much narrower applicability (i.e. only to CBs)?

database/seeds/testing/16_test_user_role.seed.js
```diff
@@ -1,7 +1,14 @@
 exports.seed = (knex) =>
   knex('user_account_access_role')
     .insert([
-      { user_account_id: 1,
-      organisation_id: 1,
-      access_role_id: 1,
-  }]);
+      { 
+        user_account_id: 1,
```
eliascodes: can we use sub-queries here to make it clearer what things are being related?

database/utils/permissions.js
```diff
@@ -0,0 +1,54 @@
+const api = require('../../src/api/v1/api.json')
+
+const permissionsForRoles = Object.values(api.routes).reduce((acc, el) => {
```
eliascodes: Can we format this better please, it's quite hard to read. Better variable names would be great too (`el`, `el2`, `acc2` 🤷🏼‍♂️).

database/utils/permissions.js
```diff
@@ -0,0 +1,54 @@
...
+const scopeToPermission = ((permission) => {
+  const [entity, permissionLevel, access] = permission.split(rx)
+  return {
+    permission_entity: entity,
+    permission_level: permissionLevel,
+    access_type: access
+  }
+})
+
+const accessRolePermissionsRows = (client) => Object.entries(permissionsForRoles).reduce((acc, el) => {
```
eliascodes: Same comment here about formatting/variable names

src/auth/permissions/__tests__/index.test.integration.ts
```diff
@@ -55,13 +58,21 @@ describe('Permisions Module', () => {
   describe('::grantExisting', () => {
     test('SUCCESS - Grant a role a permission entry that already exists', async () => {
       try {
+        const permissionId = await knex('permission').select('permission_id').where({
+          permission_entity: Resource.CONSTANTS,
+          access_type: Access.WRITE,
+          permission_level: PermissionLevel.ALL,
+        }).then((x) => x[0].permission_id);
```
eliascodes: If you only need the first, you can add `.limit(1)` onto the query. Then you can destructure the id directly.

src/auth/permissions/__tests__/index.test.integration.ts
```diff
@@ -55,13 +58,21 @@ describe('Permisions Module', () => {
...
-          access: Access.READ,
-          permissionLevel: PermissionLevel.OWN,
+          access: Access.WRITE,
+          permissionLevel: PermissionLevel.ALL,
           role: Role.VOLUNTEER });
-        expect(query).toEqual(expect.arrayContaining([{ access_role_id: 2, permission_id: 1 }]));
+        expect(query)
+          .toEqual(expect.arrayContaining([{ access_role_id: 2, permission_id: permissionId }]));
       } catch (error) {
+        console.log(error);
```
eliascodes: 🔥

src/auth/permissions/__tests__/index.test.integration.ts
```diff
@@ -26,12 +26,15 @@ describe('Permisions Module', () => {
   describe('::grantNew', () => {
     test('SUCCESS - Grant a role a permission entry that doesn\'t already exist', async () => {
       try {
+        const currentPermissions = await knex('permission').select('*');
```
eliascodes: You can select `COUNT(*)` so you don't have to do `.length`.

src/auth/permissions/__tests__/index.test.integration.ts
astroash: Knex returns an array even if you use `.limit()`

src/api/v1/permissions.json
```diff
@@ -0,0 +1,116 @@
...
+    ],
+    "VOLUNTEER": [
+      "constants-all:read",
+      "volunteer_logs-parent:write",
+      "volunteer_logs-parent:delete",
+      "user_details-own:write",
+      "user_details-own:read",
+      "volunteer_logs-child:read" 
+    ],
+    "VOLUNTEER_ADMIN": [
```
eliascodes: Shouldn't the volunteer admin have read/write over logs too?

src/api/v1/permissions.json
```diff
@@ -0,0 +1,116 @@
...
+    "VISITOR": [
+      "constants-all:read",
+      "visit_activies-own:read",
+      "visit_logs-parent:write",
+      "user_details-own:write",
+      "user_details-own:read" 
+    ],
+    "VOLUNTEER": [
+      "constants-all:read",
+      "volunteer_logs-parent:write",
```
eliascodes: Would `volunteer_logs-own:...` make more sense here? Do volunteers "own" the logs they create? Or is it the CBs?

src/api/v1/permissions.json
```diff
@@ -0,0 +1,116 @@
...
+      "user_details-own:delete",
+      "constants-all:write",
+      "constants-all:delete"
+    ],
+    "FUNDING_BODY": [
+      "constants-all:read",
+      "organisations_details-child:read",
+      "organisations_details-child:write",
+      "organisations_details-child:delete",
+      "organisations_subscriptions-child:read",
```
eliascodes: It's fine because we haven't designed it yet, but just to note there's no `organisation_subscriptions-child:write` yet.

src/auth/permissions/__tests__/index.test.integration.ts
```diff
@@ -26,12 +26,18 @@ describe('Permisions Module', () => {
   describe('::grantNew', () => {
     test('SUCCESS - Grant a role a permission entry that doesn\'t already exist', async () => {
       try {
+        const currentPermissions = await knex('permission')
+          .select('')
+          .count()
+          .then((res) => Number(res[0].count));
```
eliascodes: Up to you which you go with, just throwing out the equivalent destructuring:
```ts
const [{ count }] = await knex('permission').select().count();
```
Thought that means you have to do the type casting elsewhere. 🤷🏼‍♂️

astroash: Lol, I initially wrote it like this, but thought it was less readable 🤷🏽‍♀️

astroash:
@eliascodes I'm struggling to get the typings for `jest` working, as they have made some changes to the api. 

See issue 19783 in [DefinitelyTyped](https://github.com/DefinitelyTyped)

src/api/v1/__tests__/api.json.test.unit.ts
```diff
@@ -0,0 +1,60 @@
...
+          'isImplemented',
+          'auth',
+          'intendedFor',
+          'scope',
+          requestType,
+          'response']);
+      });
+
+      test('correct values', () => {
+        expect(typeof route.description).toEqual('string');
```
eliascodes: What's the advantage of doing it like this instead of using Joi schema?

src/auth/permissions/__tests__/index.test.integration.ts
```diff
@@ -190,12 +203,34 @@ describe('Permisions Module', () => {
...
+          permission_level: 'all',
+        },
+        {
+          access_role_id: 1,
+          access_type: 'read',
+          permission_entity: 'visit_activies',
+          permission_level: 'own',
+        },
+        {
+          access_role_id: 1, access_type: 'write',
```
eliascodes: Consistent formatting please, one attribute per line

src/models/__tests__/user.test.integration.ts
```diff
@@ -21,9 +21,9 @@ describe('User Model', () => {
 
   describe('Read', () => {
     test('get :: no arguments gets all users', async () => {
-      const users = await Users.get(knex);
+      const users = await Users.get(knex, { order: ['id', 'asc'] });
```
eliascodes: This test case is for calling `get` with no arguments (other than the client), so I'd prefer the assertion to change instead of the query here.

The assertion can become something like:
```js
expect(users).toEqual(expect.arrayContaining(expect.objectContaining({ name: 'Chell', ... })));
```

src/api/v1/__tests__/api.json.test.unit.ts
astroash: I think of `joi` as runtime validation, rather than part of a testing suite. But I guess I can write a schema and validate it during a test? Do we have any examples of using like this in the repo?

eliascodes: Yeah, there's the config tests

src/models/__tests__/user.test.integration.ts
```diff
@@ -23,16 +23,16 @@ describe('User Model', () => {
     test('get :: no arguments gets all users', async () => {
       const users = await Users.get(knex);
 
-      expect(users.length).toBe(3);
-      expect(users[0]).toEqual(expect.objectContaining({
+      expect(users.length).toBe(4);
+      expect(users).toEqual(expect.arrayContaining(expect.objectContaining({
```
eliascodes: This will need to be `expect.arrayContaining([expect.objectContaining({...`

PR #82, opened 2018-07-20
👷🏽‍♀️ Change install script for travs

astroash:
- change install script to avoid wiping node_modules cache


astroash:
Doesn't seem to speed things up

.travis.yml
```diff
@@ -21,7 +21,7 @@ services:
 
 install:
   - npm install -g codecov
-  - npm ci
+  - npm install --package-lock-only
```
eliascodes: Err, oops:
> The --package-lock-only argument will only update the package-lock.json, instead of checking node_modules and downloading dependencies.

I didn't double check what this actually did. Let's just try it with `install` again.

astroash:
Wait, I don't think this actually made a difference...

eliascodes:
Lol, oh. The build immediately after was ~3mins, I thought that was pretty good, but the one after was ~6mins, so maybe not. Basically looks like there's near enough no difference for now, so w/e.

PR #93, opened 2018-07-28
Auth rejig

### LINKED ISSUE #91: Thoughts on Roles/Permissions modules
In the process of looking at #87, I took another look at the Roles/Permissions modules and had some thoughts.

* [x] Let's make your suggested change of suffixing `Enum` on the end of enumerations
* [x] Inputs and outpus are all in terms of _database IDs_ (implementation detail) instead of _values_ (conceptually more natural). For example it would be good if I could call `Roles.fromUser` and get a value (`'VISITOR'` or `RolesEnum.VISITOR`) instead of a number that describes that value.
* [x] The `ExistsQueryResponse` type should really just be `boolean`
* [x] The `QueryResponse` type doesn't need to be a union type, because `T` in `Promise<T>` is only the type of the resolve value.

Another thing I noticed is that in the Auth strategy definition, the roles/permissions interfaces are being initialised _on each request_. I didn't catch this on review (my bad), but what I intended was for the interfaces each to be initialised once on startup, and then stored to be referenced later (much like the `knex` client is). That was kind of the whole (poorly expressed) point of that pattern; you're binding a variable to a group of functions so that you don't need to pass it in each time.

In general I still kind of like that pattern, however, looking at the code, I can see it actually reduces the possibility for re-use. Example: the `move` method is basically a combination of `userHas`, `remove` and `add`, but it's not possible to call the latter two because they need to be done as a transaction. Passing in the knex client allows us to pass in `Knex.Transaction` instances as well as `Knex` or `Knex.QueryBuilder` instances.

What do you think?

eliascodes:
Fixes #91 

The diff looks bigger than it really is because I made formatting changes that it doesn't seem to deal with very well.

#### Changes
* Use transactions to speed up integration tests ~3x
* Consolidate types for both Roles and Permissions modules
* Suffix `Enum` to all Role/Permission enumerations to prevent naming clashes
* Get rid of `Initialiser` and just pass `knex` client explicitly to functions
* Make code formatting more consistent
* Re-use code (notably `Roles.move`)
* `validate_user` function now attaches entire `User` and `Organisation` models to `auth.credentials` object instead of only IDs.


src/auth/scheme/validate_user.ts
```diff
@@ -29,26 +31,36 @@ type ValidateUser = (a: {userId: number, organisationId: number}, b: Hapi.Reques
...
-    return { isValid: false };
+
+    const [
+      user,
+      organisation,
+      userRole,
+    ] = await Promise.all([
+      Users.getOne(knex, { where: { id: userId } }),
+      Organisations.getOne(knex, { where: { id: organisationId } }),
+      Roles.fromUser(knex, { userId, organisationId }),
```
astroash: The new title of this method (`Roles.fromUser`) reads as though it return an array of associated roles. This is not the case as it grabs the first response. For clarity in calling code I think this could be renamed to something like `Roles.singleRowFromUser`.

tests/global_setup.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * This is run once before any tests run and before the framework is
+ * installed into the runtime
+ */
+const knex = require('knex');
+const { getConfig } = require('../build/config');
+const { migrate } = require('../database');
+
+
+module.exports = async () => {
```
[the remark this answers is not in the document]
eliascodes: No, but this is a _global_ setup. It runs once before any integration tests, and we know that at least one of those tests will touch the DB and the DB should be in a known state before that happens.

src/auth/scheme/validate_user.ts
```diff
@@ -16,11 +19,10 @@ const createScopeName: CreateScopeName = ({
 }) =>
   `${permission_entity}-${permission_level}:${access_type}`;
 
-
 type Credentials = {
-  userId: number,
-  organisationId?: number,
-  roleId: number
+  user: User
```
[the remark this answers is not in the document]
eliascodes: I think keeping a dozen extra attributes in memory is worth the convenience of not having to fetch the user/org models later on in pre-reqs and handlers.

src/auth/scheme/validate_user.ts
eliascodes: `singleRow` evokes the database, which is not something we should be concerned about; The module should abstract that away. How's `Roles.oneFromUser`?

tests/global_setup.js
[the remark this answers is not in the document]
astroash: My concern is that this will significantly slow down tests if we are truncating and reseeding the database when it’s not  needed. This may not be too noticeable now, but will get worse as our tests grow.

eliascodes:
> Can you also link me to the article/issue which states that transactions can be used to speed up tests.

I don't have one, I read about the technique in some forum post or something, AFAIK it's based on the technique Ruby-on-Rails uses. Running integration tests on my machine went from ~27s to ~10s.

src/auth/roles/index.ts
```diff
@@ -1,146 +1,137 @@
...
+        organisation_id: organisationId,
+        access_role_id: client('access_role')
+          .select('access_role_id')
+          .where({ access_role_name: role }),
+      });
+    const { rows } = await client.raw('SELECT EXISTS ?', [inner]);
+    return rows[0].exists;
+  },
+
+  fromUser: async (client, { userId, organisationId }) => {
```
astroash: @eliascodes can this be changed to `oneFromUser`

eliascodes: Oops, sorry, forgot. Will do.

PR #94, opened 2018-07-30
Hook

### LINKED ISSUE #59: Response post-processing
Some of the request parameters are used to change the shape of the response object (e.g. including/omitting fields, pagination, etc).

Most of this will be generic, and so should be handled by a common module. Possibly using a hapi request lifecycle event hook.

eliascodes:
~~**MERGE #93 first**~~

Fixes #59 

Adding lifecycle hooks to provide shortcuts for handlers.

Will add tests in a bit.


eliascodes:
@astroash the one uncovered branch causing this drop in coverage will only be covered once there's a handler that throws/returns a `Boom` object. I could make a fake one in a test, but frankly it seems overkill for 0.3%.

src/api/v1/hooks/onPostHandler.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * onPostHandler Lifecycle hook
+ *
+ * This hook runs immediately after the handler returns, and before
+ * any validation of the response is performed.
```
astroash: Where is validation on the response performed?

src/api/v1/index.ts
```diff
@@ -10,35 +10,51 @@
...
-      plugin: AuthJwt,
-      once: true },
-    {
-      plugin: AuthBearer,
-      once: true }]
-    );
+    /*
+     * Authentication Strategies
+     */
+    await server.register([
```
astroash: always hating on my layout...

src/api/v1/users/__tests__/index.test.integration.ts
```diff
@@ -1,23 +1,29 @@
...
 import { getConfig } from '../../../../../config';
 
 describe('API /users', () => {
+  let server: Hapi.Server;
   const config = getConfig(process.env.NODE_ENV);
   const knex = Knex(config.knex);
 
+  beforeEach(async () => {
+    server = await init(config, { routes, knex });
+    addHooks(server);
```
astroash: shouldn't this be added in `tests/server.ts`?

src/api/v1/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/*
+ * If the response data is "raw", it is wrapped appropriately.
+ * Otherwise if the response has meta-data attached to it (and
+ * is therefore correctly formatted) it is simply passed through
+ */
+export const formatResponse = (res: Hapi.ResponseObject): Response => {
+  const r: any = res.source;
+
+  if (r.hasOwnProperty('data') && r.hasOwnProperty('meta')) {
```
astroash: Do we have access to the request in this hook? If so I think it would make more sense to create metadata here (or in another hook) so that we don't have to do it individually in each handler.

tests/database.ts
```diff
@@ -1,11 +1,22 @@
...
-export const getTrx = (context: Dictionary<any>, client: Knex) =>
+
+/*
+ * Open a transaction and resolve with the transaction object.
+ * Used so that all tests within a suite can be run within a transaction
+ * (to be rolled back in an "afterEach" hook). This speeds up tests
+ * significantly compared to truncating and re-seeding the entire DB for
+ * each test
+ *
+ * Adapted from:
```
astroash: 🙏🏽 thanks for this

src/api/v1/index.ts
eliascodes: lol, can't have them inline when there's 20 of them

src/api/v1/hooks/onPostHandler.ts
eliascodes: Inside hapi

src/api/v1/users/__tests__/index.test.integration.ts
eliascodes: Fair

src/api/v1/utils.ts
eliascodes: Agreed, but at the moment the metadata consists only of pagination information (see the type definition). Once there are other things of interest that aren't handler-specific, I'm definitely up for putting them here.

src/api/v1/users/__tests__/index.test.integration.ts
eliascodes: Although I'd actually prefer to do this in #96 to avoid conflicts.

src/api/v1/utils.ts
eliascodes: One of the fields in `meta` is `total`, meaning the total number of records available to the client. How will the hook know about that?

astroash: Good point, didn't think of that. I guess it will be easier to have a helper that builds this for us in queries.

Issue #63, opened 2018-07-12
Flesh out interface tests


eliascodes:
The interface tests don't do much yet, but the API specification is (in theory) detailed enough that we can generate tests cases for each endpoint that test the following cases:

* [ ] No modifiers, full response
* [ ] Fields query, pagination, partial response
* [ ] Filter query, partial response
* [x] Malformed query, 400 response
* [x] Malformed payload, 400 response
* [x] Malformed URL parameters, 400 response
* [x] Unrecognised URL parameters, 404 response
* [x] Unauthenticated
* [x] Unauthorised(N)

We can use [fixture-factory](https://github.com/peterKaleta/fixture-factory) to generate test data.

eliascodes:
This died with the `api.json`, unfortunately.

PR #96, opened 2018-07-31
Api interface tests

### LINKED ISSUE #95: Inconsistency in API spec
https://github.com/TwinePlatform/twine-api/blob/a92ab82c47a5aaa8e8bfc551f0c43407202c7fd0/src/api/v1/api.json#L138-L146

Why does this (and other constant-routes) route have `auth: false`, but a defined `scope`? If the route is not authenticated, scope is irrelevant.

eliascodes:
**MERGE #94 FIRST**

Related to #63 
Fixes #95 

Basically I've added a few more test cases to the interface tests to cover the unauthenticated/unauthorised responses and the bad query/body responses.

One of the utility functions I added meant I could make simplifications to the `api.json.test.unit.ts` file too.

Also moved a couple of the test util files to an actual utils folder 🤷🏼‍♂️


src/api/v1/users/__tests__/index.test.integration.ts
```diff
@@ -4,7 +4,7 @@
 import * as Hapi from 'hapi';
 import * as Knex from 'knex';
 import routes from '..';
-import { init } from '../../../../../tests/server';
+import { init } from '../../../../../tests/utils/server';
 import addHooks from '../../hooks';
```
astroash: Can hooks be added in `/tests/server.ts`?

PR #98, opened 2018-08-02
Pdf & qr code stuffs

### LINKED ISSUE #91: Thoughts on Roles/Permissions modules

eliascodes:
Fixes #91 

Work done in the process of writing the visitor registration route.

### Changes
* Add constraints on user roles (one role per user, one org per user)
* Add QR code service
* Add PDF generator service


src/utils/index.ts
```diff
@@ -24,3 +26,25 @@ export const renameKeys: RenameKeys<Dictionary<any>, Dictionary<string>> =
     Object.keys(o)
       .reduce((acc, k) => assoc(map[k] || k, o[k], acc), {})
 );
+
+export const pipeToPromise = (stream: stream.Readable): Promise<Buffer[]> =>
```
astroash: curious of why this this is called `pipeToPromise`?

src/services/qrcode/__tests__/index.test.unit.ts
```diff
@@ -0,0 +1,18 @@
+import { createHmac } from 'crypto';
+import * as QRCode from '..';
+import { getConfig } from '../../../../config';
+
+
+describe('QRCode Service', () => {
+  const config = getConfig(process.env.NODE_ENV);
+
+  describe('create', () => {
```
astroash: Do we need a fail case here?

src/utils/index.ts
eliascodes: Streams have a method `pipe` which allows you to write the output of a `Readable` stream directly into a `Writable` stream. But I want to `pipe` the contents of a `Readable` stream into memory, and simply be notified (via promise resolution) when that's done.

src/services/qrcode/__tests__/index.test.unit.ts
eliascodes: I don't think I can force a failure here (without annoying the compiler). I'm just calling 3rd party functions.

src/services/pdf/__tests__/index.test.unit.ts
```diff
@@ -0,0 +1,17 @@
+import * as path from 'path';
+import { getConfig } from '../../../../config';
+import { PdfTemplateEnum, fromTemplate } from '..';
+
+
+describe('PDF Service', () => {
+  const { root } = getConfig(process.env.NODE_ENV);
+
+  describe('fromTemplate', () => {
+    test('Visitor QR Code Template', async () => {
```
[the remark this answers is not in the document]
eliascodes: Yes, probably. The only fail case I can reasonably create is by passing the wrong dictionary shape into the function.

src/utils/index.ts
astroash: Can this be renamed as `streamPipeToPromise` for clarity?

eliascodes: `pipe` is the verb here, so it should be first really. Can do `pipeStreamToPromise` if you like.

PR #102, opened 2018-08-05
[WIP] Visitor register

eliascodes:
Will probably fail on grounds of insufficient coverage.

### Changes
* Adds visitor registration endpoint
* Regenerates `permissions.json` (fixing inconsistencies after removing `constants-all:read` from auth-less endpoings)
* Use template-based email method when batch sending
* Add `exists` method to all models
* Make permissions model return values instead of ids (see #91)
* Move hash generation for QR code into Visitor model


eliascodes:
@astroash there's things I want to change, but it could do with a review anyway

astroash:
@eliascodes do you have a corresponding visitor app PR?

src/api/v1/users/register.ts
```diff
@@ -0,0 +1,153 @@
...
+      }
+      if (!(await CommunityBusinesses.exists(knex, { where: { id: payload.organisationId } }))) {
+        throw Boom.badRequest('Unrecognised organisation');
+      }
+
+      // Fetch the community business admin
+      const cb = await CommunityBusinesses.getOne(knex, {
+        where: { id: payload.organisationId, deletedAt: null },
+      });
+      const [admin] = await CbAdmins.fromOrganisation(knex, { id: payload.organisationId });
```
astroash: What's the benefit of this check?

src/api/v1/users/register.ts
```diff
@@ -0,0 +1,153 @@
...
+          role: RoleEnum.VISITOR,
+          userId: visitor.id,
+          organisationId: payload.organisationId,
+        });
+        return visitor;
+      });
+
+      // Send QR code to visitor
+      // // generate QR code data URL
+      const qrCode = await QRCode.create(visitor.qrCode);
```
astroash: Can this be done inside the transaction? If an error is thrown then the creation of the user will be rolled back.

src/api/v1/users/schema.ts
```diff
@@ -1,5 +1,45 @@
...
  * Route validation schema for User routes
  */
-export { query } from '../schema/request';
+import * as Joi from 'joi';
+
+
+export { query, gender, id } from '../schema/request';
 export { response } from '../schema/response';
+
+const currentYear = (new Date()).getFullYear();
```
astroash: Would this be better as a function? Otherwise we will forever be forever stuck in 2018...

src/auth/permissions/index.ts
```diff
@@ -148,7 +148,11 @@ const Permissions: PermissionInterface = {
         'access_role_permission',
         'permission.permission_id',
         'access_role_permission.permission_id')
-      .select()
+      .select({
```
astroash: How come you've removed `access_role_id`?

src/services/qrcode/__tests__/index.test.unit.ts
```diff
@@ -1,17 +1,11 @@
...
 describe('QRCode Service', () => {
-  const config = getConfig(process.env.NODE_ENV);
-
   describe('create', () => {
     test('Creates a signed HMAC and corresponding QR code data URL', async () => {
       const raw = 'foo';
-      const { payload, dataUrl } = await QRCode.create(raw);
-
-      expect(payload).toBe(createHmac('sha256', config.qrcode.secret).update(raw).digest('hex'));
+      const dataUrl = await QRCode.create(raw);
```
astroash: What was the hmac doing previously? Is this test still correctly named?

src/api/v1/users/register.ts
```diff
@@ -0,0 +1,153 @@
...
+      validate: {
+        payload: {
+          organisationId: id.required(),
+          name: visitorName.required(),
+          gender: gender.required(),
+          birthYear: birthYear.required(),
+          email: email.required(),
+          phoneNumber: phoneNumber.allow(''),
+          postCode: postCode.allow(''),
+          emailConsent: emailConsent.default(false),
```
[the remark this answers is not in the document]
eliascodes: The default should be that we don't assume consent to contact by e-mail/sms. This is GDPR compliant.

src/api/v1/users/register.ts
```diff
@@ -0,0 +1,153 @@
...
+    handler: async (request: RegisterRequest, h: Hapi.ResponseToolkit) => {
+      const { knex, payload, server: { app: { EmailService } } } = request;
+
+      /*
+       * Preliminaries
+       * (Can possibly eventually be removed into pre-requisites)
+       */
+      // Check user doesn't already exist
+      if (await Visitors.exists(knex, { where: { email: payload.email } })) {
+        throw Boom.conflict('User with this e-mail already registered');
```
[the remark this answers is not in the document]
eliascodes: 🤷🏼‍♂️

Ultimately no difference in behaviour. Want me to switch?

src/api/v1/users/register.ts
eliascodes: I need the email of the associated admin

src/auth/permissions/index.ts
eliascodes: It's not useful. We should deal with the value it describes instead.

src/services/qrcode/__tests__/index.test.unit.ts
eliascodes: No, the test name should be changed, thanks.

src/api/v1/users/register.ts
eliascodes: Possibly. I'm not sure what the deal is with doing non-DB operations inside a transaction. I'll look into it.

eliascodes: Although in fact, really the only critical ops for this endpoint end at line 102. If the transaction completes successfully we are happy, since the rest of the operations can be repeated without much bother (e.g. the CB admin can request that the code is e-mailed to the visitor again).

Maybe if the latter half of the handler throws at any stage, we can return a [202](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success) response, which doesn't quite fit, but is the closest I can find.

src/api/v1/users/register.ts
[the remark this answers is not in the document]
astroash: True, I was getting this confused with requesting consent to store peoples details. On that note I wonder if we need to keep a record of this consent.

src/api/v1/users/register.ts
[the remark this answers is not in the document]
astroash: Yes please.

src/api/v1/users/register.ts
astroash: Your latter suggestion will involve changes to the client so that there is notification of the error. Is there a corresponding `twine-visitor` PR?

My preference would be for new users to not be able to sign up unless their qrcode has been sent, otherwise you end up with the edge case where they're not allowed to sign up again because of email constraints and have to talk to an admin to get their code.

PR #114, opened 2018-08-21
Fix production seeds and other DB build errors

eliascodes:
Even more things that we missed(!)

I've added the DB builds to the travis config so that we know if something's broken in future. All the DB URLS point to the test DB provided by travis.

I've now also removed the redundant `/api` URL prefix, and added a utility function that we can use to add API tokens to the DB via the existing DB CLI.


astroash:
@eliascodes tests failing

PR #124, opened 2018-08-24
[wip] Route visit activities

astroash:


eliascodes:
@astroash dunno if you've already accounted for this, but it would be useful to be able to do something like `GET /blah?day=today` or `GET /blah?day=monday` to filter activities only for those days.

src/api/v1/api.json
```diff
@@ -347,6 +347,84 @@
           "response": null
         }
       },
+      "/:id/visit_activities": {
+        "GET": {
+          "description": "Retrieve all visit activities for an organistion",
+          "isImplemented": true,
+          "auth": true,
+          "intendedFor": ["TWINE_ADMIN", "FUNDING_BODY", "ORG_ADMIN", "VISITOR"],
```
eliascodes: So there's an issue with how these kinds of routes in the `api.json` get translated into role/permissions relations by `database/utils/permissions.js`. Roles like `VISITOR` end up with `xxxx-child:zzzz` permissions, which aren't appropriate (it should only have either `xxxx-own:zzzz` or `xxxx-parent:zzzz` depending on the resource), similar for `ORG_ADMIN`.

I "fixed" this in my PR by separating `/community-businesses/:id` routes (which have `child` scopes) from `/community-businesses/me` routes (which have `own` scopes). IMO we should do a similar thing here, which will prevent the wrong scopes being assigned to `ORG_ADMIN`, though I don't think it will solve the `VISITOR` case, since it will either use an `/:id` route or a `/me` route, and the `permissions.js` file currently has no way of differentiating which of the entries in the `scope` array should be assigned to which role.

src/api/v1/community_businesses/__tests__/visit_activites.test.integration.ts
```diff
@@ -0,0 +1,129 @@
...
+  describe('DELETE', () => {
+    test(':: successfully add a new activity', async () => {
+      const res = await server.inject({
+        method: 'DELETE',
+        url: '/v1/community-businesses/1/visit_activities/1',
+        credentials: { scope: ['visit_activities-own:delete'] },
+      });
+
+      expect(res.statusCode).toBe(200);
+      expect((<any> res.result).result.deletedAt).toBeTruthy();
```
eliascodes: I think this test should make a subsequent request to `GET` to check that the deleted activity is no longer returned.

src/api/v1/community_businesses/schema.ts
```diff
@@ -2,10 +2,45 @@
...
+  wednesday: Joi.boolean(),
+  thursday: Joi.boolean(),
+  friday: Joi.boolean(),
+  saturday: Joi.boolean(),
+  sunday: Joi.boolean(),
+};
+
+export const id = Joi.number().positive().required();
+
+export const meOrId = Joi.number().positive().allow('me').required();
```
eliascodes: No action, but just FYI, `const meOrId = id.allow('me').required();` would also work.

src/api/v1/community_businesses/visit_activities.ts
```diff
@@ -0,0 +1,130 @@
...
+import { Day } from '../../../types/internal';
+
+interface GetRequest extends Hapi.Request {
+  query: {
+    day: Day
+  };
+}
+
+interface PostOrPutRequest extends Hapi.Request {
+  payload: Partial<VisitActivity>;
```
eliascodes: Is this right? The schema for the two requests are different.

src/api/v1/community_businesses/visit_activities.ts
```diff
@@ -0,0 +1,130 @@
...
+      ],
+      validate: {
+        query: visitActivitiesGetQuery,
+        params: { communityBusinessId: meOrId },
+      },
+      response: { schema: response },
+    },
+    handler: async (request: GetRequest, h: Hapi.ResponseToolkit) => {
+      const { knex, pre: { communityBusiness }, query: { day: _day = null } } = request;
+      const day = _day === 'today'
```
eliascodes: If we treat `'today'` as a different case from all other values, and end up casting it to an explicit weekday value, surely it shouldn't be a part of the `Day` union type?

src/api/v1/community_businesses/visit_activities.ts
```diff
@@ -0,0 +1,130 @@
...
+    path: '/community-businesses/{communityBusinessId}/visit_activities',
+    options: {
+      description: 'Retrieve all visit activities for a community business',
+      auth: {
+        strategy: 'standard',
+        access: {
+          scope: [
+            'visit_activities-own:read',
+            'visit_activities-child:read',
+            'visit_activties-parent:read',
```
eliascodes: There's no differentiation between any of these scopes in the handler/pre-reqs. If I only have the `own` scope, I can hit this endpoint for _any_ CB and the handler will still give me the data.

src/models/community_business.ts
```diff
@@ -272,5 +278,104 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+      category: 'visit_activity_category.visit_activity_category_name',
+      monday: 'monday',
+      tuesday: 'tuesday',
+      wednesday: 'wednesday',
+      thursday: 'thursday',
+      friday: 'friday',
+      saturday: 'saturday',
+      sunday: 'sunday',
+      createdAt: 'created_at',
+      modified_at: 'modified_at',
```
eliascodes: camel case

src/models/community_business.ts
```diff
@@ -272,5 +278,104 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+    })
+      .where({ organisation_id: o.id, deleted_at: null });
```
eliascodes: Inconsistent indentation

src/models/community_business.ts
```diff
@@ -272,5 +278,104 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
 
     return query;
   },
+
+  async getVisitActivities (client: Knex, o: CommunityBusiness, d?: Day) {
+    const baseQuery = client('visit_activity')
+    .innerJoin(
+      'visit_activity_category',
+      'visit_activity_category.visit_activity_category_id',
+      'visit_activity.visit_activity_id')
```
eliascodes: this is wrong key to join on

src/models/community_business.ts
```diff
@@ -272,5 +278,104 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+      'tuesday',
+      'wednesday',
+      'thursday',
+      'friday',
+      'saturday',
+      'sunday',
+      'created_at AS createdAt',
+      'modified_at as modifiedAt',
+    ]);
+    return res || null;
```
eliascodes: If the query returns an empty array, doesn't that mean the insert has failed? Are you sure you want to just return `null` in that case? Also this is inconsistent with the return type in the function signature `Promise<VisitActivity>`. This function has the return type `Promise<VisitActivity | null>`

src/models/community_business.ts
```diff
@@ -272,5 +278,104 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+  },
+
+  async updateVisitActivity (client: Knex, v: Partial<VisitActivity>) {
+    const [res] = await client('visit_activity')
+    .innerJoin(
+      'visit_activity_category',
+      'visit_activity_category.visit_activity_category_id',
+      'visit_activity.visit_activity_id')
```
eliascodes: wrong key to join on

src/models/community_business.ts
```diff
@@ -272,5 +278,104 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+      .where({ visit_activity_id: v.id, deleted_at: null })
```
eliascodes: inconsistent indentation

src/api/v1/api.json
astroash: @eliascodes aren't we trashing the `VISITOR` scope?

astroash: @eliascodes aren't we trashing the concept of `VISITOR` for now?

eliascodes: The use of `VISITOR` was just an example, same goes for `ORG_ADMIN`

astroash: This should work if we drop `VISITOR` and replace `:/id/` -> `/me`. This would leave `ORG_ADMIN` as the only user type to access `/me/` so it can have `xxxx-own:zzzz`.

`TWINE_ADMIN` and `FUNDING_BODY` will use the `/:id/` route with `xxx-child:zzzz` with the appropriate prerequisites.

src/api/v1/community_businesses/__tests__/visit_activites.test.integration.ts
astroash: _done_

src/api/v1/community_businesses/visit_activities.ts
astroash: _done_

src/api/v1/community_businesses/__tests__/visit_activites.test.integration.ts
```diff
@@ -0,0 +1,152 @@
...
+        credentials: {
+          user: 1 ,
+          organisation: 1 ,
+          scope: ['visit_activities-own:delete'],
+        },
+      });
+
+      expect(res.statusCode).toBe(200);
+      expect((<any> res.result).result.deletedAt).toBeTruthy();
+      const doubleCheck = await CommunityBusinesses.getVisitActivities(server.app.knex, { id: 1 });
```
eliascodes: This is fine, but I actually meant to make a new `GET` request and check the response. The latter is more decoupled from the implementation of the handler logic, and operates on the same level of abstraction as the functionality under test (i.e. you're testing from the perspective of a consumer of the HTTP API, who wouldn't have access to lower level abstractions -- in this case the models).

This is just an FYI though, doesn't need to be done.

src/models/community_business.ts
```diff
@@ -279,5 +285,104 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+        'tuesday',
+        'wednesday',
+        'thursday',
+        'friday',
+        'saturday',
+        'sunday',
+        'created_at AS createdAt',
+        'modified_at as modifiedAt',
+      ]);
+    return res || null;
```
eliascodes: Where did we land on this being inconsistent with the type declaration (and possibly papering over a failed insert)?

src/models/community_business.ts
```diff
@@ -279,5 +285,104 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+        category: 'visit_activity_category.visit_activity_category_name',
+        monday: 'monday',
+        tuesday: 'tuesday',
+        wednesday: 'wednesday',
+        thursday: 'thursday',
+        friday: 'friday',
+        saturday: 'saturday',
+        sunday: 'sunday',
+        createdAt: 'created_at',
+        modified_at: 'modified_at',
```
eliascodes: Still missing camelCase here

src/api/v1/permissions.json
```diff
@@ -47,6 +47,7 @@
       "organisations_details-child:write",
       "organisations_feedback-child:read",
       "organisations_outreach-child:read",
+      "organisations_outreach-own:read",
```
eliascodes: Odd that `TWINE_ADMIN` has `own` permissions for organisation stuff, no?

src/api/v1/permissions.json
```diff
@@ -116,6 +116,7 @@
     "VOLUNTEER_ADMIN": [
       "constants-all:read",
       "organisations_outreach-child:read",
```
eliascodes: Some more strange permissions here and directly below. Would have expected `organisations_outreach-own:read` instead. No action in this PR, I'm just noting so I can make an issue.

src/api/v1/community_businesses/__tests__/visit_activites.test.integration.ts
```diff
@@ -0,0 +1,152 @@
...
+    server = await init(config);
+  });
+
+  afterAll(async () => {
+    await server.shutdown(true);
+  });
+
+  describe('GET', () => {
+    test(':: successfully gets all activities for a cb', async () => {
+      const res = await (<any> server).inject({
```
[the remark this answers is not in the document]
astroash: It complains about the credentials otherwise. I tried to extend the `Hapi.Server` type but couldn't get it working, n thought it was of less importance since it's for a test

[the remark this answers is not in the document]
eliascodes: The reason it was complaining about credentials is because the credentials object you're giving it is the wrong type (see my [comment below](https://github.com/TwinePlatform/twine-api/pull/124#discussion_r214087158))

src/api/v1/community_businesses/schema.ts
```diff
@@ -2,10 +2,44 @@
...
+    'sunday',
+  ]),
+};
+
+export const visitActivitiesPostPayload = {
+  name: Joi.string().required(),
+  category: Joi.string().required(),
+};
+
+export const visitActivitiesPutPayload = {
```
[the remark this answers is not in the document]
astroash: I've left the functionality in, but the visitor app currently does not support this. I can imagine this being of use in the future, if an activity is accidentally created with the wrong category.

src/api/v1/community_businesses/visit_activities.ts
```diff
@@ -0,0 +1,166 @@
...
+    options: {
+      description: 'Retrieve all visit activities for a community business',
+      auth: {
+        strategy: 'standard',
+        access: {
+          scope: ['visit_activities-own:delete'],
+        },
+      },
+      validate: {
+        payload: Joi.any().valid(null),
```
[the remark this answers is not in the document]
astroash: Yes,  but `payload: Joi.any().forbidden()` means that they key payload is forbidden rather than just it's value, so this fails any request sent to this endpoint

src/api/v1/community_businesses/schema.ts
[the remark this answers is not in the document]
eliascodes: OK, let's leave it as is for now then

eliascodes:
Don't care about minor coverage drop right now.

PR #135, opened 2018-09-02
Update outdated dependencies

eliascodes:
Does what it says on the tin


package.json
```diff
@@ -32,8 +32,8 @@
     "migrate:make": "node ./bin/db.js migrate:make",
     "migrate:teardown": "node ./bin/db.js migrate:teardown",
     "migrate:teardown:test": "NODE_ENV=testing node ./bin/db.js migrate:teardown",
-    "migrate:reset": "npm run migrate:teardown && npm run migrate:latest && npm run seed:run",
-    "migrate:reset:test": "npm run migrate:teardown:test && npm run migrate:latest:test && npm run seed:run:test",
+    "migrate:reset": "npm run migrate:teardown && npm run migrate:latest",
```
astroash: Why's this script been changed, I use it a lot. If this is to fit better with it's name, can you change it's name back to `db:reset`, and add this as a new one.

astroash:
@eliascodes is this meant to be merged after #134?

package.json
eliascodes: See #134, migration 15 is now running the seeds

PR #137, opened 2018-09-04
[WIP] Visitors get

astroash:


src/models/__tests__/visitor.test.integration.ts
```diff
@@ -49,11 +49,16 @@ describe('Visitor model', () => {
       expect(exists).toBe(false);
     });
 
-    test('fromOrganisation :: gets all visitors at organisation', async () => {
```
eliascodes: Was the `fromOrganisation` test deleted in the rebase? We should probably retrieve that.

src/api/v1/community_businesses/visitors.ts
```diff
@@ -0,0 +1,93 @@
...
+        modelQuery.where = mergeDeepRight(
+          modelQuery.where || {},
+          { gender: filter.gender }
+        );
+      }
+
+      const visitors = await (visits
+        ? Visitors.getWithVisits(
+            knex,
+            (<Pick<CommunityBusiness, 'id'>> pick(['id'], communityBusiness)),
```
eliascodes: See comment below on arg type.

src/models/visitor.ts
```diff
@@ -114,8 +126,105 @@ export const Visitors: UserCollection & CustomMethods = {
...
+  async serialise (user: Partial<User>) {
+    const strippedUser = omit(['password', 'qrCode'], user);
+    const serialisedUser = user.qrCode
+      ? assoc('qrCode', await QRCode.create(user.qrCode), strippedUser)
+      : strippedUser;
+
+    return serialisedUser;
+  },
+
+  async getWithVisits (client: Knex, c: Pick<CommunityBusiness, 'id'>, q: ModelQuery<User> = {}) {
```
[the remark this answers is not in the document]
astroash: It seemed to make more sense for this method to work with just org id rather than an entire cb object, as that's all that's needed.

src/models/__tests__/visitor.test.integration.ts
astroash: Hehe oh dear...

src/models/visitor.ts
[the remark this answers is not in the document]
eliascodes: From the perspective of a consumer of the models/collections API, I don't care what part of the models are being used, nor how the queries are formed. The model objects are my units of abstraction, and I should be able to work with them as naturally as possible.

Issue #83, opened 2018-07-20
Route Epic [Stage 2 and 3]


astroash:
- [x]  constants (#84)

### organisations
- [x] [`GET /community-businesses`](https://github.com/TwinePlatform/twine-api/blob/staging/src/api/v1/community_businesses/get.ts)
- [x] [`GET /community-businesses/:id/questions/:qid/cls`](https://github.com/TwinePlatform/twine-api/blob/staging/src/api/v1/community_businesses/questions.ts)
- [x] [`POST /community-businesses/:id/feedback`](https://github.com/TwinePlatform/twine-api/blob/staging/src/api/v1/community_businesses/feedback.ts)
- [x] [`GET /community-businesses/:id/feedback`](https://github.com/TwinePlatform/twine-api/blob/staging/src/api/v1/community_businesses/feedback.ts)
- [x] [`GET /community-businesses/:id/feedback/aggregates`](https://github.com/TwinePlatform/twine-api/blob/staging/src/api/v1/community_businesses/feedback.ts)
- [x] `/community-businesses/:id`
- [x] `GET | PUT | POST | DELETE /community-businesses/:id/visit-activities` TwinePlatform/twine-visitor#476
- [x] `POST /community-businesses/:id/visit-logs`
- [x] `GET /community-businesses/me/visit-logs` TwinePlatform/twine-visitor#482
- [x] `GET /community-businesses/me/visit-logs/aggregates`
- [x] `GET /community-businesses/:id/visitors`
- [x] `GET /community-businesses/:id/visitors/:id` #152 
- [x] `PUT /community-businesses/:id/visitors/:id`
- [x] `POST /community-businesses/me/visitors/:id/emails` #151 
- [x] `PUT /community-businesses/:id`
- [x] `POST /community-businesses/me/volunteer-logs` #175 
- [x] `GET /community-businesses/me/volunteer-logs/summary` @eliascodes 
- [x] `POST /community-businesses/me/volunteer-logs/sync` @eliascodes 

### users
- [x] [`GET /users`](https://github.com/TwinePlatform/twine-api/blob/staging/src/api/v1/users/get.ts)
- [x] [`GET /users/me`](https://github.com/TwinePlatform/twine-api/blob/staging/src/api/v1/users/get.ts)
- [x] `PUT /users/me`
- [x] [`POST /users/register/visitor`](https://github.com/TwinePlatform/twine-api/blob/staging/src/api/v1/users/register.ts)
[...]

PR #146, opened 2018-09-10
Put users

eliascodes:
**MERGE #145 FIRST**
Related to #83 

### Changes
* Add `PUT /users/me`


eliascodes:
...You know I think I did the wrong endpoint.

I did `PUT /users/me/` and `PUT /users/:id`. I probably should have done `PUT /users/me` and `PUT /community-businesses/:id/visitors/:id`.

I can add the latter 🤷🏼‍♂️

src/api/v1/community_businesses/visitors.ts
```diff
@@ -80,11 +104,91 @@ const routes: Hapi.ServerRoute[] = [
...
+      if (!isChild) {
+        return Boom.forbidden('Insufficient permission to access this resource');
+      }
+
+      const changeset = { ...payload };
+
+      const [user] = await Visitors.fromCommunityBusiness(
+        knex,
+        communityBusiness,
+        { where: { id: Number(userId) } }
```
astroash: This can be parsed as a number by joi validation instead of manually in the route. If we never want this value as a string this may be a better approach.

src/api/v1/prerequisites/is_child_user.ts
```diff
@@ -0,0 +1,56 @@
+/*
+ * "user_details-child" route pre-requisite
+ *
+ * Determines if the authenticated user is trying to access an user
+ * which is actually a "child" entity.
+ *
+ * Conditions under which this is true for user X attemptying to access:
```
astroash: Is this prereq intended for accessing unserialised user data? If so I think we should mention that FUNDING_BODY_ADMIN's are unauthenticated for this reason.

If not then we should add a note to add their use case at a later date

src/api/v1/community_businesses/visitors.ts
```diff
@@ -80,11 +104,91 @@ const routes: Hapi.ServerRoute[] = [
...
-        : Visitors.get(knex, modelQuery));
+        : Visitors.fromCommunityBusiness(knex, communityBusiness, modelQuery));
 
       return Promise.all(visitors.map(Visitors.serialise));
     },
   },
+
+  {
+    method: 'PUT',
+    path: '/community-businesses/{organisationId}/visitors/{userId}',
```
astroash: Is it worth mentioning that this is a duplicate route?

src/api/v1/community_businesses/visitors.ts
eliascodes: It is parsed by Joi, the problem is that the hapi typings expect all path parameters to be strings, so this is just to satisfy the compiler.

src/api/v1/community_businesses/__tests__/visitors.test.integration.ts
```diff
@@ -87,5 +106,143 @@ describe('API /users/visitors', () => {
...
+          user,
+          organisation,
+          scope: ['user_details-child:write'],
+        },
+      });
+
+      expect(res.statusCode).toBe(400);
+    });
+
+    test('idempotency', async () => {
```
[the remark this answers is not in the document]
eliascodes: ?

src/api/v1/prerequisites/is_child_user.ts
eliascodes: What do you mean by "accessing unserialised user data"

astroash: User data that allows the user to be identified. Due to us not collecting gdpr consent for  funding bodies they cannot access personal info data about users.

src/api/v1/community_businesses/visitors.ts
astroash: Or even refactoring out the handler to avoid duplication

src/api/v1/community_businesses/__tests__/visitors.test.integration.ts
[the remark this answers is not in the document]
astroash: Not the most verbose test description

src/api/v1/community_businesses/visitors.ts
astroash: Typescript = the worst

src/api/v1/prerequisites/is_child_user.ts
eliascodes: OK. But they will be allowed to view aggregate or anonymised data. We may need to add new permissions for that. Maybe `user_aggregates-child:read` or similar. I'll add a note to the description here.

Also, for reference: [serialisation](https://en.wikipedia.org/wiki/Serialization)

PR #153, opened 2018-09-14
Pwd reset

eliascodes:
Related to #83 

### Changes
* Add `POST /users/password/reset` route
* Merge test suites of `POST /users/password/forgot` and `POST /users/password/reset`.


src/models/user.ts
```diff
@@ -239,31 +241,61 @@ export const Users: UserCollection & CustomMethods = {
 
   async createPasswordResetToken (client: Knex, u: Partial<User>) {
     const twoDaysFromToday = moment().add(2, 'days').toISOString();
-    const token = randomBytes(64).toString('base64');
+    const token = randomBytes(32).toString('hex');
```
astroash: why's the length been reduced?

eliascodes: Bytes are not the same as characters. 64 bytes encoded in base 64 gives 88 characters. 32 bytes encoded in hex gives 64 characters.

PR #156, opened 2018-09-17
🌟 add lastWeek field for visit-log/aggregates

astroash:
- update visit test data to be added in relation to today rather than weekday
- add lastWeek option to CBs.getVisitLogAggregates


src/models/community_business.ts
```diff
@@ -602,6 +603,27 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
             } , {});
           return { age };
         }),
+      lastWeek: applyQueryModifiers(client('visit')
```
eliascodes: Why is this an aggregate of its own? Why not just use the `whereBetween` options on existing aggregates.

astroash: Each aggregate is counting visits split up into certain categories, with this one splitting up the result into days of the week. This query could use the `CBs.getVisitLogs` method with a whereBetween, but would then need post processing display the data correctly ie:
```
  lastWeek: {
    '2018-09-05': 4,
    '2018-09-06': 2,
    '2018-09-07': 0 ...
  }
```

For readability, this method is defining the query itself.

src/models/community_business.ts
```diff
@@ -602,6 +603,27 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+        .select({ createdAt: 'visit.created_at' })
+        .whereRaw('visit.created_at >= CURRENT_DATE - INTERVAL \'7 day\'')
+        .innerJoin('visit_activity', 'visit_activity.visit_activity_id', 'visit.visit_activity_id')
```
eliascodes: What are all these joins for? Doesn't look like they're used.

src/models/community_business.ts
```diff
@@ -602,6 +603,27 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+        .innerJoin('user_account', 'user_account.user_account_id', 'visit.user_account_id')
+        .innerJoin('gender', 'gender.gender_id', 'user_account.gender_id'), queryMatchOnColumnNames)
+        .then((data) => {
+          const lastWeekDatesObject = Array
```
eliascodes: I don't think it's necessary to construct the object with zeros here if you just overwrite it below.

src/models/community_business.ts
```diff
@@ -602,6 +603,27 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
...
+            .apply(null, { length: 7 })
+            .map((x: null, i: number) =>
+              today.clone().subtract(i + 1, 'days').format('DD-MM-YYYY'))
+            .reverse()
+            .reduce((o: Dictionary<number>, key: string) => ({ ...o, [key]: 0 }), {});
+
+          const lastWeek = data.reduce((acc: Dictionary<number>, visit: any) => {
+            const visitDateKey = moment(visit.createdAt).format('DD-MM-YYYY');
+            acc[visitDateKey] = acc[visitDateKey] + 1;
```
eliascodes: `acc[visitDateKey] = (acc[visitDateKey] || 0) + 1;` makes lines 613-618 unnecessary.

src/models/community_business.ts
```diff
@@ -602,6 +603,27 @@ export const CommunityBusinesses: CommunityBusinessCollection & CustomMethods =
             } , {});
           return { age };
         }),
+      lastWeek: applyQueryModifiers(client('visit')
+        .select({ createdAt: 'visit.created_at' })
+        .whereRaw('visit.created_at >= CURRENT_DATE - INTERVAL \'7 day\'')
+        .innerJoin('visit_activity', 'visit_activity.visit_activity_id', 'visit.visit_activity_id')
+        .innerJoin('user_account', 'user_account.user_account_id', 'visit.user_account_id')
+        .innerJoin('gender', 'gender.gender_id', 'user_account.gender_id'), queryMatchOnColumnNames)
+        .then((data) => {
```
eliascodes: Seems like this can be typed as `Pick<VisitEvent, 'createdAt'>[]`, which means your typing on the reduce below can be tighter too.

src/models/community_business.ts
astroash: These are used if there is a query passed in, see line L510-L519

Issue #157, opened 2018-09-18
🔐 incorrect scope for creating a user


astroash:
Currently `user_details-child:write` requires full access mode. However, for the visitor app `ORG_ADMINS` need to be able to create/sign up a new user whilst in restricted mode. 

Relates TwinePlatform/twine-visitor#516

PR #158, opened 2018-09-18
🗃 update user_details-child:write to restricted access

astroash:
**Changes**:
- update user_details-child:write to restricted access

**Fixes** #157


database/migrations/sql/20180918_18_make_user_details_permission_restricted.sql
```diff
@@ -0,0 +1,21 @@
+/*
+ * Make user_details-child:write restricted access
+ * for creating new visitors on the visitor app
+ */
+
+ UPDATE access_role_permission
```
eliascodes: This needs to be an insert, permissions are queried either on "full" or "restricted", not "full" + "restricted".

astroash: Was that meant to say 'or'?

If not I'm unsure what you mean.

Issue #149, opened 2018-09-12
Detailed validation error response


eliascodes:
Should ensure any request validation errors are detailed in the response payload to allow accurate error reporting in clients (much like the visitor app currently assumes).

astroash:
For future reference, joi error messages are no longer passed to the client with hapi 17,  generic 400 is returned. For custom messages `failAction` needs to be used.

https://github.com/hapijs/hapi/issues/3706#issuecomment-349637461
https://medium.com/@csakis/hapi-joi-error-handling-with-custom-messages-1f544e2b1489

astroash:
Custom validation error messages have been added to `users/password/reset` using the above method. Visitor app routes that are still setup for custom validation messages, and need to implement `validation.failAction`:

- [ ] `/users/password/forgot`
- [ ] `/users/login/admin`
- [ ] `/users/register/visitor`

eliascodes:
@astroash this is done now right?

PR #165, opened 2018-09-20
Validation errors

astroash:
Changes:
- add validation error formatter to hooks
- fix validation on password/reset route  …
- strengthen regex for strong password
- fix matching validation for passwordConfirm
- fix payload values to confirmPassword -> passwordConfirm
- add failaction for validation error formatting
- return 401 for token error
- add tests to check validation errors on route

Relates #149


src/api/v1/users/password.ts
```diff
@@ -74,7 +75,18 @@ const routes: Hapi.ServerRoute[] = [
...
           password: passwordSchema,
-          confirmPassword: Joi.any().only(Joi.ref('password')),
+          passwordConfirm: Joi.string().required().valid(Joi.ref('password'))
+            .options({ language: { any: {
+              required: 'confirm password is required',
+              allowOnly: 'must match password',
+            } } }),
+        },
+        failAction: (request, h, err) => {
+          if ((<JoiBoomError> err).details[0].message
```
eliascodes: IMO this is still a 400. The route has `auth: false`, anyone can access it, it just happens to require a token.

src/api/v1/users/password.ts
```diff
@@ -74,7 +75,18 @@ const routes: Hapi.ServerRoute[] = [
         payload: {
           token: Joi.string().length(64),
           password: passwordSchema,
-          confirmPassword: Joi.any().only(Joi.ref('password')),
+          passwordConfirm: Joi.string().required().valid(Joi.ref('password'))
+            .options({ language: { any: {
+              required: 'confirm password is required',
```
eliascodes: I think `password confirmation is required` makes more sense here?

src/api/v1/users/password.ts
```diff
@@ -90,7 +102,7 @@ const routes: Hapi.ServerRoute[] = [
         user = await Users.fromPasswordResetToken(knex, token);
       } catch (error) {
         request.log('warning', error);
-        return Boom.badRequest('Invalid token');
+        return Boom.unauthorized('Invalid token. Reset password again.');
```
eliascodes: Again, think this is a `400`

src/api/v1/utils.ts
```diff
@@ -5,6 +5,28 @@ import * as Hapi from 'hapi';
...
 import { ApiResponse } from './schema/response';
 
+export interface JoiBoomError extends Boom {
+  details: {message: string}[];
+}
+
+/*
+ * formatJoiValidation allows joi errors thrown from
+ * failAction to be reformatted before sending response.
+ * Note joi throws an error on the first value that fails
```
eliascodes: This is configurable. Do we want to include all fields that fail validation?

src/api/v1/utils.ts
```diff
@@ -5,6 +5,28 @@ import * as Hapi from 'hapi';
...
+
+export const formatJoiValidation = ({ output: { payload }, details }: JoiBoomError)
+  : ApiResponse => {
+  const message = details[0].message.split('"');
+  return {
+    error: {
+      statusCode: payload.statusCode,
+      type: payload.error,
+      message: details[0].message,
+      validation: { [message[1]]: message[2].trim() },
```
eliascodes: Check out this for another way to do this that protects against cases where `"` appears in the error message:

https://github.com/TwinePlatform/twine-visitor/blob/1c8da775d5373e853e9678abac311b87b847dd75/server/shared/middleware/validation.js#L72-L80

src/api/v1/utils.ts
```diff
@@ -5,6 +5,28 @@ import * as Hapi from 'hapi';
 import * as Boom from 'boom';
 import { ApiResponse } from './schema/response';
 
+export interface JoiBoomError extends Boom {
+  details: {message: string}[];
```
eliascodes: If this is a declaration that should actually be in `@types/boom` but is missing, we should use declaration merging instead of creating a new type.

Something like: 
```ts
declare module 'boom' {
  class Boom extends Error {
    details: { message: string }[];
  }
}
```
I don't know if you actually need to attach it to the class or to the namespace though. Take a look at `@types/boom/index.d.ts`.

Also, because I'd like to eventually replace all our custom declaration fixes with PRs to the real typings, it'd be nice to keep them in one place (which I haven't been doing)

src/api/v1/utils.ts
```diff
@@ -5,6 +5,28 @@ import * as Hapi from 'hapi';
...
+}
+
+/*
+ * formatJoiValidation allows joi errors thrown from
+ * failAction to be reformatted before sending response.
+ * Note joi throws an error on the first value that fails
+ * validation, therefore 'details' will only have 1 element.
+ */
+
+export const formatJoiValidation = ({ output: { payload }, details }: JoiBoomError)
```
eliascodes: I think this function can probably be folded into the other one once we use declaration merging  as in the other comment.

src/api/v1/users/password.ts
astroash: The token is being used to authorise the password reset. We are sending back `400` if there is an error with a payload that is supplied by the user, it's validating their input. IMO `401` better suits token errors as it's authentication that we have created and send to the user in an email rather than the standard auth of a cookie.

src/api/v1/utils.ts
astroash: This may be useful for signup routes where we are taking a bigger number of fields from the user, but for now I think we should leave it as it is.

src/api/v1/users/schema.ts
```diff
@@ -16,8 +16,9 @@ export const userName =
   Joi.string()
     .min(3)
     .max(100)
-    .regex(/[^\w\s\d]/, { name: 'alphanumeric', invert: true })
-    .lowercase();
+    .regex(/^[a-zA-Z]{2,}\s?[a-zA-z]*['\-]?[a-zA-Z]*['\- ]?([a-zA-Z]{1,})?/, { name: 'alpha' })
```
eliascodes: Rather than come up with ever more specific regexes to handle every edge case you can think of, why not just use a permissive regex so you don't need to worry about the edge cases you have inevitably missed.

There is really no need to tightly constrain user names. TBH, this is why I wasn't really a fan of lowercasing names in the database. It's purely a developer convenience.

I'm not going to say we should deal with this now though. We should ultimately deal with it by using the `citext` postgres module or a functional index to support case insensitive searches.

No action.

src/api/v1/utils.ts
```diff
@@ -7,21 +7,52 @@ import { ApiResponse } from './schema/response';
 
 
 /*
- * We only choose to re-format the boom response
+ * Boom error type does not include key details
+ * added by Joi failAction validation method
+ */
+export interface BoomWithValidation extends Boom {
```
eliascodes: This can be `type ... = Boom<any> & Joi.ValidationError;` and it'll cover the other fields.

PR #168, opened 2018-09-25
Model typings, security headers & other

eliascodes:
@astroash these are the consolidation of the model typings I was talking about. Also some other changes that I caught along the way.

### Changes
* Consolidate model type declarations in `models/types.ts`, do not re-declare method types
* Set security headers & CORS settings
* Update docs on directory structure
* Prevent `DeepPartial` from breaking arrays by using conditional typing 🙈


config/types.ts
```diff
@@ -17,7 +17,7 @@ type WebConfig = {
     cors: Hapi.RouteOptionsCors
     security: Hapi.RouteOptionsSecureObject | boolean
   },
-  tls: null | { key: string, cert: string }
+  tls?: null | { key: string, cert: string }
```
astroash: Haven't you removed this from the config?

eliascodes: Yeah, we'll never use it. It's only useful if we deploy on a VPS or something else where there's no routing front-end that can handle things like SSL for us. We're on heroku now, and if we ever migrate it'll be to a similar cloud provider so I can't see it being needed.

src/models/index.ts
```diff
@@ -5,12 +5,17 @@ export { Organisations } from './organisation';
...
   LinkedFeedback,
+  VisitActivity,
+  VisitEvent,
+  LinkedVisitEvent,
+  Subscription,
   GenderEnum,
+  DisabilityEnum,
+  EthnicityEnum,
   ModelQuery,
   CommonTimestamps,
```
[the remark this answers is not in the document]
eliascodes: Yes, as a convenience, see api/v1/types

Issue #170, opened 2018-09-27
Authentication routes


eliascodes:
Visitor app (cookie):
* Should only support cb admins logging in

Volunteer app (header):
* Should support volunteers and volunteer admins logging in
* Possibly should also support cb admins logging in

Dashboard (cookie):
* Should only support cb admins logging in

URLs in a RESTful API should describe resources or actions, and ideally we would aim not to over-specialise endpoints to serve only a single client application (we have done this in the past, but largely out of pragmatism; we didn't know enough about use cases to generalise).

Suggestion: use query or payload parameters to constrain behaviour of route:
`/users/login?restrict={role}&type={cookie | header}`

This is ultimately not that different from having `/users/login/admin`, `/users/login/volunteer`, etc, but doesn't require several routes to describe one semantic, and is more explicit in what it means at face value. Multiple allowed roles are supported through arrays in `?restrict[]={role1}&restrict[]={role2}`.

Not that fussed whether these should be query params or payload params.

astroash:
@eliascodes would you be able to add something to the documentation regarding this choice so we can close this issue?

astroash:
@eliascodes bump

eliascodes:
I think this should be de-prioritised until there is more clarity on #110

astroash:
@eliascodes can this be closed?

Issue #167, opened 2018-09-25
Authentication should support Auth header


eliascodes:
In order to support the native app, the API needs to inspect request `Authorization` headers as well as cookies when looking for the JWT.

eliascodes:
It does!

PR #171, opened 2018-09-27
User login

eliascodes:
Related to #170, #167 

Update to the login route as outlined in #170 

### Changes
* `Roles.userHas` supports list of roles
* Rename `/users/login/admin` to `users/login`, update logic to support all user types
* Update handler logic to support returning token in payload instead of as a cookie


src/auth/roles/index.ts
```diff
@@ -85,15 +85,18 @@ const Roles: RolesInterface = {
   },
 
   userHas: async (client, { role, userId, organisationId }) => {
+    const sub = Array.isArray(role)
+      ? client('access_role') .select('access_role_id') .whereIn('access_role_name', role)
```
astroash: Why are you using `whereIn`, are you expecting `role` to be an array?

eliascodes: Yes, hence `Array.isArray`.

astroash: Can it be changed to `roles`?

src/api/v1/users/__tests__/login.test.integration.ts
```diff
@@ -0,0 +1,166 @@
+import * as Hapi from 'hapi';
+import * as Knex from 'knex';
+import * as JWT from 'jsonwebtoken';
+import { init } from '../../../../server';
+import { getConfig } from '../../../../../config';
+import { getCookie } from '../../../../utils';
+import { getTrx } from '../../../../../tests/utils/database';
+
+
+describe('POST /users/login/admin', () => {
```
astroash: 🔥 'POST /users/login/**admin**'

src/auth/roles/index.ts
eliascodes: It's not always an array, hence the ternary.

astroash: Thought it read better as plural, but feel free to leave.

PR #175, opened 2018-10-01
volunteer logs endpoints

eliascodes:
### Changes
* Add `GET /users/volunteers/me/volunteer-logs`
* Add `POST /community-businesses/me/volunteer-logs`
* Add `GET /users/volunteers/me/volunteer-logs/:id`
* Add `PUT /users/volunteers/me/volunteer-logs/:id`
* Add `DELETE /users/volunteers/me/volunteer-logs/:id`


src/api/v1/users/volunteer_logs.ts
```diff
@@ -0,0 +1,298 @@
...
+  query,
+  ApiRequestQuery,
+  id,
+} from './schema';
+import { VolunteerLogs, CommunityBusiness, VolunteerLog, CommonTimestamps } from '../../../models';
+import { Omit } from '../../../types/internal';
+import { getCommunityBusiness } from '../prerequisites';
+
+
+interface GetMyVolunteerLogsRequest extends Hapi.Request {
```
astroash: Can this go in `api/v1/types.ts`?

src/api/v1/users/volunteer_logs.ts
```diff
@@ -0,0 +1,298 @@
...
+  query: ApiRequestQuery & {
+    since: string
+    until: string
+  };
+  pre: {
+    communityBusiness: CommunityBusiness
+  };
+}
+
+interface PostMyVolunteerLogsRequest extends Hapi.Request {
```
astroash: 🚚

src/api/v1/users/volunteer_logs.ts
```diff
@@ -0,0 +1,298 @@
...
+  payload: Pick<VolunteerLog, 'activity' | 'duration' | 'startedAt'> & {
+    organisationId?: number
+  };
+  pre: {
+    communityBusiness: CommunityBusiness
+  };
+}
+
+interface GetVolunteerLogRequest extends Hapi.Request {
```
astroash: 🚚

src/api/v1/users/volunteer_logs.ts
```diff
@@ -0,0 +1,298 @@
...
+  query: { fields: (keyof VolunteerLog)[] };
+  params: { logId: string };
+  pre: {
+    communityBusiness: CommunityBusiness
+  };
+}
+
+interface PutMyVolunteerLogRequest extends Hapi.Request {
```
astroash: 🚚

src/api/v1/users/volunteer_logs.ts
```diff
@@ -0,0 +1,298 @@
...
+    communityBusiness: CommunityBusiness
+  };
+}
+
+
+const routes: Hapi.ServerRoute[] = [
+
+  {
+    method: 'GET',
+    path: '/users/me/volunteer-logs',
```
astroash: `/users/me/volunteer-logs` -> `/users/volunteers/me/volunteer-logs`

🚚 mv `api/v1/users/...` -> `api/v1/users/volunteers/...`

src/api/v1/users/volunteer_logs.ts
```diff
@@ -0,0 +1,298 @@
...
+        { since, until, limit: query.limit, offset: query.offset }
+      );
+
+      return Promise.all(logs.map(VolunteerLogs.serialise));
+    },
+  },
+
+  {
+    method: 'POST',
+    path: '/users/me/volunteer-logs',
```
astroash: 🚚 `/users/me/volunteer-logs` -> `/community-businesses/me/volunteer-logs`

src/api/v1/users/volunteer_logs.ts
```diff
@@ -0,0 +1,298 @@
...
+      description: 'Create volunteer log for own user',
+      auth: {
+        strategy: 'standard',
+        access: {
+          scope: ['volunteer_logs-parent:write'],
+        },
+      },
+      validate: {
+        payload: {
+          organisationId: id,
```
astroash: 🔥

src/api/v1/users/volunteer_logs.ts
```diff
@@ -0,0 +1,298 @@
...
+          return Boom.badRequest('Invalid activity');
+        }
+        throw error;
+      }
+    },
+  },
+
+  {
+    method: 'GET',
+    path: '/users/me/volunteer-logs/{logId}',
```
astroash: 🚚 `/users/me/volunteer-logs/:id` -> `/community-businesses/me/volunteer-logs/:id`

eliascodes: I don't think this one should move. If we can access the entire collection through `GET /users/volunteers/me/volunteer-logs` it follows that we can also access individual resources through `GET /users/volunteers/me/volunteer-logs/:id`

astroash: IMO I don't think that any logs should be accessible under the `/users` endpoint. Logs are owned by cbs and should be accessed as such.

astroash: On second look, I don't think that any logs should be accessible under the /users endpoint. Logs are owned by cbs and should be accessed as such.

src/api/v1/__tests__/api_interface.test.integration.ts
```diff
@@ -77,7 +77,7 @@ const testSpec = Object.entries(collapseUrls(APISpecification.routes, '/v1'))
     /*
      * Bad request tests
      */
-    if (routeSpec.query !== null) {
+    if (routeSpec.query && routeSpec.query !== null) {
```
astroash: Why are you checking if it exists, then checking it doesn't equal `null`?

src/api/v1/api.json
```diff
@@ -885,6 +890,23 @@
             "totalVolunteerHours": "Number",
             "volunteerLogs": ["types.volunteerLog"]
           }
+        },
+        "POST": {
+          "description": "Create volunteer logs for own community business",
```
astroash: How do you know which user this log is attached to?

src/api/v1/users/volunteers/__tests__/volunteer_logs.test.integration.ts
```diff
@@ -0,0 +1,361 @@
...
+            organisationId: organisation.id,
+            userId: user.id,
+            activity: 'Outdoor and practical work',
+            duration: { hours: 5 },
+          }),
+        ]),
+      });
+    });
+
+    test('can get no logs for non-volunteer', async () => {
```
astroash: So, under our new error handling plan...

Would the method in this handler throw an error, and this route return a 404?

eliascodes: Yeah this was unchanged from before the URL change. When it was `/users/me/volunteer-logs`, a `200 []` response made sense because `/users/me` is an existing resource, but one that has no `volunteer-logs` associated to it.

Now that the URL has changed, `/users/volunteers` is a collection that does not include `me` if that `me` is not a volunteer, so `/users/volunteers/me` (and all sub-routes) should be a 404.

As far as how this 404 is generated, we could either put a check in the model method and/or we could check the value of `role` in the auth credentials.

src/api/v1/users/volunteers/volunteer_logs.ts
```diff
@@ -0,0 +1,201 @@
...
+        pre: { communityBusiness },
+        params: { logId },
+      } = request;
+
+      const log = await VolunteerLogs.getOne(knex, { where: {
+        id: Number(logId),
+        userId: user.id,
+        organisationId: communityBusiness.id,
+        deletedAt: null,
+      }});
```
astroash: Out of curiosity, have you turned off tslint?

src/api/v1/api.json
eliascodes: Can admins create logs for other volunteers? If so, we can add an optional `userId` field in the payload.

src/api/v1/users/volunteers/volunteer_logs.ts
eliascodes: No

astroash: Weird, on my local it adds spaces between closing brackets.

eliascodes: Mine too but not in this case, must be an edge case in the parser?

src/api/v1/api.json
eliascodes: erm, I've encountered an issue:

`VOLUNTEER` and `VOLUNTEER_ADMIN` users are both given the `volunteer_logs-parent:write` scope. `ORG_ADMIN` users are given the `volunteer_logs-own:write` scope.

We want `VOLUNTEER_ADMIN` and `ORG_ADMIN` users to be able to add logs for other volunteers at their organisation. But, as mentioned, `VOLUNTEER_ADMIN` and `VOLUNTEER` users have the same scope, so it's not possible to prevent `VOLUNTEER` users from being able to add logs for other users purely through scopes.

The kinda gross solution to this is to simply check the user role in the handler. The reason I think this is kinda gross is because we have been trying to avoid explicit reference to roles in handlers where possible. I'm happy to do it for now, but if you have scope-related idea to solve this, I'd be interested in hearing it.

astroash: My initial response it to change the concept of `logs` to be owned by the user who they’re related to. Then we would have `volunteer_logs-sibling` for `VOLUNTEER_ADMIN` and `volunteer_logs-own` for `VOLUNTEER`s

But I’ll have a think if there’s another solution.

eliascodes: I think this needs more thought, so perhaps worth continuing with the bad solution and addressing this separately. Can open an issue.

astroash: @eliascodes can you comment all areas this fix is implemented with a `//TO BE REPLACED`

Then happy to merge this

PR #182, opened 2018-10-04
Register volunteers and friends

astroash:
~~**MERGE #185 first**~~

Changes:
- add `Volunteers.adminCodeIsValid`
- add `Volunteers.addWithRole` which will add either a volunteer_admin or volunteer type user depending on args
- ~~add `CommunityBusinesses.byRegion`~~
- add `GET regions/:id/community-businesses`
- add `POST users/register/volunteers`
- move `POST users/register/visitors` to own file

_see commits for more details_

Notes | Disclaimers | Don't hate me's:
- `GET regions/:id/community-businesses` resides in `/v1/regions` whereas `GET regions` is in `v1/constants`. It did not feel appropriate to put the new handler in the constants folder as the data will change, but I have no strong feelings on this
- `.adminCodeIsValid` method is separated out for clarity, but could be rolled into 
`.addWithRole`


src/api/v1/users/register/__tests__/visitors.test.integration.ts
```diff
@@ -0,0 +1,136 @@
+import * as Hapi from 'hapi';
```
eliascodes: The diff thinks this is a new file, but I'm assuming it's just been moved and renamed. Did you make any changes to the file?

src/api/v1/users/register/volunteers.ts
```diff
@@ -0,0 +1,93 @@
...
+    method: 'POST',
+    path: '/users/register/volunteers',
+    options: {
+      description: 'Retreive list of all users',
+      auth: false,
+      validate: {
+        payload: {
+          organisationId: id.required(),
+          role: Joi.alternatives(RoleEnum.VOLUNTEER, RoleEnum.VOLUNTEER_ADMIN),
+          adminCode: Joi.string().regex(/^[0-9]{5}$/),
```
eliascodes: This doesn't support the existing admin codes. Also lets leave ourselves a little leeway to support more complex codes.

src/api/v1/users/register/volunteers.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (await Users.exists(knex, { where: { email } })) {
+        return Boom.conflict('User with this e-mail already registered');
+      }
+      const communityBusiness = await CommunityBusinesses
+        .getOne(knex, { where: { id: organisationId, deletedAt: null } });
+      // check organisation exists
+      if (!communityBusiness) return Boom.badRequest('Unrecognised organisation');
+
+      /*
+       * Registation
```
eliascodes: typo

src/api/v1/users/register/volunteers.ts
```diff
@@ -0,0 +1,93 @@
...
+        volunteer = await Volunteers.addWithRole(knex, payload, role, communityBusiness, adminCode);
+
+      } catch ({ message }) {
+        if (message === 'Invalid volunteer admin code') {
+          return Boom.unauthorized(message);
+        }
+      }
+
+      // register login event
+      await Volunteers.recordLogin(knex, volunteer);
```
eliascodes: Maybe we shouldn't do this here if volunteers can be created by others; it doesn't really count as a login anymore, I think.

src/api/v1/regions/community_businesses.ts
```diff
@@ -0,0 +1,21 @@
+import * as Hapi from 'hapi';
+import { CommunityBusinesses } from '../../../models';
+import { response } from '../schema/response';
+
+
+export default [
+  {
+    method: 'GET',
+    path: '/regions/community-businesses',
```
eliascodes: This isn't what I meant. If you look at the message I sent, I used the URL `/regions/:id/community-businesses`, not `/regions/community-businesses`. The semantic being expressed is "for region :id, what community businesses are registered?". 

Is there a reason you want all community businesses/regions in one go?

src/models/volunteer.ts
```diff
@@ -102,9 +131,20 @@ export const Volunteers: VolunteerCollection = {
...
     return omit(['password', 'qrCode'], user);
   },
+
+  async adminCodeIsValid (client, cb, code) {
+    const [row] = await client('volunteer_admin_code')
+      .where({
+        code,
+        organisation_id: cb.id,
+      });
+    return row ? true : false;
```
eliascodes: `Boolean(row)` avoids a branch here

src/models/__tests__/community_business.test.integration.ts
```diff
@@ -296,4 +297,50 @@ describe('Community Business Model', () => {
...
+        region: RegionEnum.LONDON,
+        sector: SectorEnum.COMMUNITY_HUB,
+      });
+      await CommunityBusinesses.add(trx, {
+        name: 'Okamis Fields',
+        region: RegionEnum.LONDON,
+        sector: SectorEnum.ENVIRONMENT,
+      });
+      /*
+       * NB: due to sequences not resetting during transaction rollbacks this test
```
eliascodes: Then just don't assert against the IDs. Use `expect.objectContaining`.

src/api/v1/users/register/volunteers.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (role === RoleEnum.VOLUNTEER_ADMIN && !adminCode) {
+        return Boom.badRequest('Missing volunteer admin code');
+      }
+      let volunteer;
+
+      try {
+        volunteer = await Volunteers.addWithRole(knex, payload, role, communityBusiness, adminCode);
+
+      } catch ({ message }) {
+        if (message === 'Invalid volunteer admin code') {
```
eliascodes: so if it's any other message we just do nothing and continue? Gives a runtime error because `volunteer` will be undefined.

src/api/v1/users/register/volunteers.ts
```diff
@@ -0,0 +1,93 @@
...
+       * Registation
+       */
+
+      if (role === RoleEnum.VOLUNTEER_ADMIN && !adminCode) {
+        return Boom.badRequest('Missing volunteer admin code');
+      }
+      let volunteer;
+
+      try {
+        volunteer = await Volunteers.addWithRole(knex, payload, role, communityBusiness, adminCode);
```
eliascodes: There's not so much code left that you can't just move it all into the `try` block. Then you won't need the `let volunteer`.

src/api/v1/users/register/volunteers.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (!communityBusiness) return Boom.badRequest('Unrecognised organisation');
+
+      /*
+       * Registation
+       */
+
+      if (role === RoleEnum.VOLUNTEER_ADMIN && !adminCode) {
+        return Boom.badRequest('Missing volunteer admin code');
+      }
+      let volunteer;
```
eliascodes: How did you get away with not giving this a type declaration?

src/api/v1/regions/community_businesses.ts
astroash: Yes, it is needed for the volunteer app.

src/api/v1/users/register/volunteers.ts
astroash: Can you clarify?

src/api/v1/users/register/__tests__/visitors.test.integration.ts
astroash: Think its a product of my workflow. I initially added `users/register/volunteers` in the same file, then decided to separate for clarity.

src/models/__tests__/community_business.test.integration.ts
eliascodes:
```ts
expect(cbByRegion).toEqual({
  London: [
    expect.objectContaining({ name: 'Aperture Science' }),
    expect.objectContaining({ name: 'Black Mesa Research' }),
    expect.objectContaining({ name: 'Okamis Fields' }),
    expect.objectContaining({ name: 'Zeldas Dungeon' }),
  ]
});
```

Doesn't seem like much extra. Or even:
```ts
expect(cbByRegion).toEqual({
  London: [
    { name: 'Aperture Science' },
    { name: 'Black Mesa Research' },
    { name: 'Okamis Fields' },
    { name: 'Zeldas Dungeon' },
  ].map(expect.objectContaining)
});
```

src/api/v1/users/register/__tests__/visitors.test.integration.ts
eliascodes: OK, but have you actually changed any of the code in this file?

src/api/v1/users/register/volunteers.ts
eliascodes: Existing admin codes are alphanumeric and longer than 5 characters (I think 8). The bit about more complex codes was about allowing symbols and such, but actually that's probably annoying for users who have to enter these codes, so ignore that part.

src/api/v1/users/register/__tests__/visitors.test.integration.ts
eliascodes: OK, but have you actually changed any of the code in this file? I just want to know whether I need to pay attention to this or not.

src/api/v1/regions/community_businesses.ts
eliascodes: Is there a PR I can look at? I'm interested to see what it's needed for.

src/models/__tests__/community_business.test.integration.ts
astroash: @eliascodes I know how to implement what you were suggesting. I wrote this test as it is because it explicitly shows that the results are ordered in alphabetically rather than by id. It is also gives a more readable return value for when you are looking back at tests.

As this is test code, not application code, I do not think we need to worry about the small case when one of us wants to run this test in isolation.

src/api/v1/regions/community_businesses.ts
astroash: `www/js/controllers/register-controller.js` in https://github.com/TwinePlatform/twine-app/pull/21

You can see the UX on the register page of the app.

src/api/v1/users/register/__tests__/visitors.test.integration.ts
astroash: Not to my recollection (apart from the endpoint name)

src/api/v1/regions/community_businesses.ts
eliascodes: Originally it worked fetching for a single region, like I imagined. Now we're grabbing all the data in one go.

I don't think it's more performant this way tbh. If you imagine that in the most likely use case, a volunteer will know what region they're in, so fetching by region will result in less data transferred.

astroash:
@eliascodes see latest commit. 

POSTGRES Error:
```
error: column "community_business" of relation "community_business" does not exist
```

SQL:
```sql
with "new_organisation" as 
  (insert into "organisation" ("_360_giving_id", "organisation_name") 
    values (?, ?) returning *) 
    
insert into "community_business" ("address_1", "community_business"."community_business_region_id", "community_business"."community_business_sector_id", "logo_url", "organisation_id", "post_code", "town_city", "turnover_band") 
values (
  ?, 
  (select "community_business_region_id" from "community_business_region" where "region_name" = ?), 
  (select "community_business_sector_id" from "community_business_sector" where "sector_name" = ?), 
  ?, 
  (select "organisation_id" from "new_organisation"), 
  ?, 
  ?, 
  ?
) returning "organisation_id"',
      returning: 'organisation_id
```

I believe the error is with the dot notation in the SQL values. I was considering `evolv`ing these columnames back, but I wanted to check that I am using existing methods as intended.

src/api/v1/regions/community_businesses.ts
```diff
@@ -0,0 +1,38 @@
...
+import { pick, map } from 'ramda';
+import { CommunityBusinesses } from '../../../models';
+import { response } from '../schema/response';
+
+
+export default [
+  {
+    method: 'GET',
+    path: '/regions/{regionId}/community-businesses',
+    options: {
```
eliascodes: Let's add some validation on the `regionId` parameter? There's a `id` schema knocking about somewhere (I think in schema/request).

src/api/v1/regions/community_businesses.ts
```diff
@@ -0,0 +1,38 @@
...
+      const { server: { app: { knex } }, params: { regionId } } = request;
+
+      const [{ regionName } = { regionName: null }] = await knex('community_business_region')
+        .select({ regionName: 'region_name' })
+        .where({ community_business_region_id: regionId })
+        .limit(1);
+
+      if (!regionName) { return Boom.notFound('Region cannot be found'); }
+
+      const regions = await CommunityBusinesses.get(knex, {
```
eliascodes: I guess you had regions on the brain but these are community businesses, no?

PR #183, opened 2018-10-09
Org summary

eliascodes:
~~**MERGE #179 FIRST**~~

Related to #83 

### Changes
* Adds `GET /community-businesses/me/volunteer-logs/summary`
* Adds `POST /community-businesses/me/volunteer-logs/sync`


database/migrations/sql/20181009_20_add_unique_start_time_constraint_volunteer_hours_log.sql
```diff
@@ -0,0 +1,8 @@
+/*
+ * Add unique start time constraint on volunteer_hours_log
+ *
+ * User should not be able to register multiple logs that have identical
+ * start times.
+ */
+ALTER TABLE volunteer_hours_log
```
astroash: @eliascodes why is this being added as a migration?

src/api/v1/community_businesses/volunteer_logs.ts
```diff
@@ -99,6 +99,134 @@ const routes: Hapi.ServerRoute[] = [
...
+          return Boom.badRequest('Duplicate logs in payload, check start times');
+        }
+        throw error;
+      }
+    },
+  },
+
+  {
+    method: 'GET',
+    path: '/community-businesses/me/volunteer-logs/summary',
```
astroash: Is there a difference between `summary` and `aggregates` endpoints? From what I've seen we have `cb../summary` and `users../aggregates` returning similar data

src/api/v1/community_businesses/volunteer_logs.ts
```diff
@@ -99,6 +99,134 @@ const routes: Hapi.ServerRoute[] = [
...
+      } = request;
+
+      /*
+       * This manual checking of scopes is necessary because
+       * hapi.js scopes specification isn't capable of capturing
+       * complex scope rules e.g ((x AND y) OR (a AND b)).
+       *
+       * See: https://hapijs.com/api#-routeoptionsauthaccessscope
+       */
+      const hasParentScopes =
```
astroash: Can this all be removed when I change the permissions?

database/migrations/sql/20181009_20_add_unique_start_time_constraint_volunteer_hours_log.sql
eliascodes: I think this was before we decided we could re-do the DB. I'll move it into the original migrations.

src/api/v1/community_businesses/volunteer_logs.ts
eliascodes: I don't think so, because these checks aren't about which permissions are associated to which roles, but about specifying more complex scope requirements than hapi can natively support.

When we change `parent` to `child`, we'll still need to check that they have both 'org_details' and 'volunteer_logs', unless we decide that the 'org_details' scope is unnecessary.

src/api/v1/community_businesses/volunteer_logs.ts
eliascodes: I think I only used `summary` because that's what the old endpoint was called. I can switch to `aggregates`, no prob.

astroash: I think I would prefer both of these endpoints to be named `/summary`. Then we can keep `/aggregates` for when we return dynamic aggregated data for the super admin dashboard.

src/api/v1/community_businesses/volunteer_logs.ts
astroash: Why are there so many scopes? I tried looking at the `api.json` but this route isn't in there 🙃

eliascodes: lol, oops.

So my reasoning was that `org_details` was necessary here because we're querying "how many volunteers are there at the CB?", which doesn't seem like a `volunteer_logs` related question.

We _could_ overlook that in this instance, and then I'd only need `volunteer_logs` scopes and could get rid of the extra logic. 🙈

astroash: would `volunteer_logs-sibling:read` cover this use case? To me that says I have permission to read the logs of other users at this org.

eliascodes: But normal volunteers don't have that scope, right? From what I remember, there's no indication in the native app code that this feature is only for admins (but maybe we can just say that it will be?)

astroash: I'm not following. Where on the volunteer app is this endpoint used?

eliascodes: Only on the dashboard controller

astroash: I'd be tempted to stick with `organisations_details-parent:read`. I think of logs as data related to a user, this info is annoymised and summarised making it have more in common with generic cb data, than the concept of volunteer_logs.

eliascodes: OK, so just to be clear, we're saying we'll stick with the 4 scopes and the additional logic? I'm happy with that, I just want to know whether there's any remaining action here.

astroash: No, I was suggesting dropping the `volunteer_logs` scopes.

eliascodes:
Think I've addressed everything, LMK if not!

PR #185, opened 2018-10-09
#databaseprotest

### LINKED ISSUE #184: Volunteer admin codes
Related to TwinePlatform/twine-app#22, TwinePlatform/data-sync#18

These admin codes have been mistakenly modeled as one-time codes. In fact they are long-lived tokens.

This requires the following changes to the DB model:
* `single_use_token` renamed to `token`
* `single_use_token.expires_at` made nullable
* `volunteer_admin_one_time_code` renamed to `volunteer_admin_code`
* `user_id` foreign key replaced with `organisation_id` foreign key

These tokens should also probably be hashed before storage.

astroash:
🗃 update volunteer_admin_code table
- rename table
- link to organisation
- add code & timestamp columns

relates #184
(not fixing as it doesn't address hashing)


database/migrations/sql/20180604_04_create_organisations.sql
```diff
@@ -75,6 +75,17 @@ CREATE TABLE community_business (
   CONSTRAINT community_business_to_community_business_region_fk FOREIGN KEY (community_business_region_id) REFERENCES community_business_region
 );
 
+CREATE TABLE volunteer_admin_code (
+  volunteer_admin_code_id SERIAL NOT NULL UNIQUE,
+  organisation_id                INT NOT NULL,
+  code                           INT NOT NULL,
```
eliascodes: `VARCHAR`

database/migrations/sql/20180604_04_create_organisations.sql
```diff
@@ -75,6 +75,17 @@ CREATE TABLE community_business (
   CONSTRAINT community_business_to_community_business_region_fk FOREIGN KEY (community_business_region_id) REFERENCES community_business_region
 );
 
+CREATE TABLE volunteer_admin_code (
+  volunteer_admin_code_id SERIAL NOT NULL UNIQUE,
+  organisation_id                INT NOT NULL,
```
eliascodes: Do we want to support multiple admin codes per organisation? If not this should be `UNIQUE`

astroash: Previously I thought these were 5 digit numerical values and didn't want to restrict ourselves to 99999. But I was wrong, so this can go back in 😊

astroash: Oh wait, I though this was on the `code` column. Do we want this to be unique as well?

PR #209, opened 2018-10-25
Logs routes

astroash:
**Changes:**
- add missing FUNDING_BODY RoleEnum
- add GET /visit-logs
   - only implemented for TWINE_ADMIN
   - with tests
   - add comments for incomplete route
- add GET /volunteer-logs
   - implemented for TWINE_ADMIN
   - add TODO notes
   - update VolunteerLogs.get to allow optional fields
   - with tests

**NOTE**:
- new routes are added to `/api/v1/user_logs` directory as `/logs` is in `.gitignore`. Let me know if you can think of a more appropriate directory name


src/models/volunteer_log.ts
```diff
@@ -88,7 +92,14 @@ export const VolunteerLogs: VolunteerLogCollection = {
           'volunteer_activity',
           'volunteer_activity.volunteer_activity_id',
           'volunteer_hours_log.volunteer_activity_id')
-        .select(query.fields ? pick(query.fields, ModelToColumn) : ModelToColumn),
+        .innerJoin(
```
eliascodes: Obviously this is fine for now. Ideally in future I'd like to think of a way to get rid of it again though 🙊

astroash: Yeah, I would also like to separate out the types at some point so that additional fields are not optional on main type declarations.

PR #212, opened 2018-10-31
Update password forgot/reset

astroash:
**Changes:**
- add dev scripts for locally adding users/reset tokens
- add `redirect` to forgot payload
- add new template id for postmark
- remove `Users.fromPasswordResetToken`
- add `Users.usePasswordResetToken`
- invalidate all old tokens when creating a new one
- maybe fix those pesky tests


src/models/user.ts
```diff
@@ -265,34 +276,41 @@ export const Users: UserCollection = {
...
+    // check token hash matches supplied token
+    const isValid = await compare(token, match.token);
+
+    if (!isValid) throw new Error('Token not recognised');
+
+    // update token row as used
     await client('single_use_token')
-      .update({ used_at: new Date() })
-      .where({ single_use_token_id: tokenRow.single_use_token_id });
+      .update({ used_at: moment().toISOString() })
```
eliascodes: No action, but is there a reason for `moment` here? There's no real difference between `new Date()` and `moment()` if they're only being used for the ISO86001 output. 🤷🏼‍♂️

PR #219, opened 2018-11-06
Update routes for cbs

astroash:
**Changes:**
- update GET /cbs to use queries
- fix PUT /cbs turnoverBand validation
- fix GET /visit-logs to include logs without a category
- add _360GivingId to PUT cb/:id || me


src/api/v1/community_businesses/put.ts
```diff
@@ -33,7 +33,8 @@ export default [
           address2: Joi.string().min(1),
           townCity: Joi.string().min(1),
           postCode: Joi.string().min(6).max(10),
-          turnoverBand: Joi.string().min(6).max(11),
+          turnoverBand: Joi.string().min(5).max(11),
+          _360GivingId: Joi.string().min(5).max(11),
```
eliascodes: I'm not sure this is long enough in general; we don't know for definite how long the identifier strings are (charity and company numbers are like 7-8 digits, but there might be others used, and the prefixes can be longer than 6 characters (example: http://org-id.guide/list/XI-BRIDGE)).

I think we can just put this at something reasonably high (e.g. 255), and not worry about it yet, since we don't have a real validation schema for 360 IDs.

PR #220, opened 2018-11-07
🗃 add timestamps to consts

### LINKED ISSUE #191: 🤔 Adding timestamp columns to database constants
Currently we are grabbing a number of constants from the database such as regions, sectors and genders. These tables do not have timestamp columns which means that all constant routes in the api will return every db row.

If in the future any of these constants change we will need to hard delete rows, which will not allow us to support 'legacy' data which may be linked to old constants. 

I suggest we add the standard `created_at`, `modified_at`, `deleted_at` columns for all of these tables for flexibility of future use cases

astroash:
☝🏽 

fixes #191


database/migrations/sql/20180604_08_create_visits.sql
```diff
@@ -5,6 +5,9 @@
 CREATE TABLE visit_activity_category (
   visit_activity_category_id   SERIAL NOT NULL UNIQUE,
   visit_activity_category_name VARCHAR(255) NOT NULL,
+  created_at                   TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  modified_at                  TIMESTAMP WITH TIME ZONE,
```
eliascodes: All of these `modified_at` columns need to have triggers added so they auto-update.

PR #222, opened 2018-11-07
Citext

### LINKED ISSUE #169: Use citex instead of lower-casing values in DB
Sparked by: https://github.com/TwinePlatform/twine-api/pull/165#discussion_r220148797

We shouldn't be transforming source names only as a developer convenience. Instead we can use the `citext` postgres module to do indexed case-insensitive searches.

astroash:
fixes #169 

**Changes:**
- add citext for user added data
- rename database reset script
- remove joi parsing to lower/upper case
- fix tests


database/migrations/sql/20180604_08_create_visits.sql
```diff
@@ -28,6 +28,7 @@ CREATE TABLE visit_activity (
 
   CONSTRAINT visit_activity_pk                            PRIMARY KEY (visit_activity_id),
   CONSTRAINT visit_activity_to_community_business_fk      FOREIGN KEY (organisation_id)            REFERENCES organisation ON DELETE CASCADE,
+  CONSTRAINT visit_activity_name_length                             CHECK (char_length(visit_activity_name) <= 100),
```
eliascodes: 255?

PR #228, opened 2018-11-12
🐛 qr codes not stored hashed

### LINKED ISSUE #227: 🐛 visitor qr codes not being hashed - make it so everywhere
🙈

astroash:
- unhash test seed qrcode
- update handler to search via qrcode and send appropriate error messages
- (☝🏽 both search and POST visit-log)
- update tests
- 🔥 findAsync as it's no longer used in the codebase

fixes #227


src/api/v1/community_businesses/visit_logs.ts
```diff
@@ -58,15 +54,19 @@ const routes: Hapi.ServerRoute[] = [
...
     handler: async (request: VisitorSearchRequest, h) => {
-      const { payload: { qrCode, userId, visitActivityId }, server: { app: { knex } } } = request;
+      const { payload: { userId, visitActivityId }, server: { app: { knex } } } = request;
       const communityBusiness = <CommunityBusiness> request.pre.communityBusiness;
 
-      const visitors = await Visitors.fromCommunityBusiness(knex, communityBusiness);
+      const visitor = await Visitors.getOne(knex, { where: { id: userId } });
 
-      const visitor = await findAsync(visitors, (v: User) => bcrypt.compare(qrCode, v.qrCode));
+      const isRegisteredVolunteerAtCb = await Roles.userHas(knex, {
```
eliascodes: `isRegisteredVisitorAtCB`

src/api/v1/community_businesses/visit_logs.ts
```diff
@@ -58,15 +54,19 @@ const routes: Hapi.ServerRoute[] = [
...
+        userId,
+        organisationId: communityBusiness.id,
+        role: RoleEnum.VISITOR,
+      });
 
-      if (!visitor || visitor.id !== userId) {
-        return Boom.badRequest('QR code invalid');
+      if (!isRegisteredVolunteerAtCb) {
+        return Boom.forbidden('Visitor is not registered at Community Business');
```
eliascodes: IMO this is still `400`. `403` indicates that the user-agent making the request is forbidden from accessing the resource. In this case, the user-agent is permitted to access the resource, but must submit an appropriate user ID.

src/api/v1/users/visitors/search.ts
```diff
@@ -40,13 +37,10 @@ const routes: Hapi.ServerRoute[] = [
...
       const {
         payload: { qrCode },
-        pre: { communityBusiness },
         server: { app: { knex } },
       } = request;
 
-      const visitors = await Visitors.fromCommunityBusiness(knex, communityBusiness);
-
-      const visitor = await findAsync(visitors, (v: User) => compare(qrCode, v.qrCode));
+      const visitor = await Visitors.getOne(knex, { where: { qrCode } });
```
eliascodes: Do we want to be able to search across all CBs here? Genuine question.

astroash: Yes, I guess, as the endpoint is nested under `users`. The app probably wants to hit a cbs/me/users though...

src/api/v1/community_businesses/visit_logs.ts
astroash: Which resource are you describing here?

I was imagining the specific user outlined in the payload as the resource owner of logs. Other CBs are forbidden from adding logs for them.

eliascodes: In a REST(ish) API, the resource is described by the URL: the visit logs registered against the user-agents' community business. Anyone who has `visit_logs-child:write` scope is authorised to add or edit logs (in this case the CB admin).

src/api/v1/users/visitors/search.ts
eliascodes: I guess let's leave it as is for now.

src/api/v1/community_businesses/visit_logs.ts
astroash: I think we're better describing our resources by how our permissions are defined. Switching between resource ownership in permissions and resource definitions in urls will end up with inconsistencies.

Endpoints have been simplified due to readability eg this endpoint should ideally be `cbs/users/visit-logs`. Whereas, IMO, resource ownership in permissions has made less concessions _(but maybe that's because I wrote them 🙃)_

eliascodes: Alright, well I'm not sure I quite agree, but also don't want to get into a big discussion over this right now, so let's leave it as is.

astroash: Make an issue, I'm happy to be persuaded otherwise 😊

eliascodes:
https://github.com/TwinePlatform/twine-api/pull/228#discussion_r233115116 is the remaining change.

astroash:
@eliascodes const has been renamed by ammending the original commit.

PR #236, opened 2018-11-20
Config cors

eliascodes:
### Changes
* Set CORS origin via env vars.
* Pull common config into default


astroash:
@eliascodes apologies, I know I said otherwise earlier but...

Can you explain the reasoning for moving CORS origin URLs to env variables? 

Personally, I am a fan of config files as they serve as one place for collating all settings needed to spin up the server in different environments. This makes it simpler to clone/setup the server from scratch, and acts as a clear comparison for the config implemented for different environments. 

From my current understanding the only benefit of using env variables for cors origins is to allow us to insert separate variables to the staging branch. This functionality can easily be reproduced by having a `config.staging.ts` file which is evaluated if `NODE_ENV=staging`. IMO this would be more beneficial as:
- it will allows us to implement other specific config settings needed for staging deployment as the needs arise, rather than having to refactor more settings into env variables. 
- all _non sensitive_ config is held in one place

I did some reading around the topic, and couldn't find any reference of other benefits. If I am missing something LMK.

https://gist.github.com/telent/9742059
https://12factor.net/config
https://laracasts.com/discuss/channels/laravel/env-vs-config-files

eliascodes:
Strictly speaking, the system of dev/test/prod environments based on a `NODE_ENV` switch is purely a developer convenience, meant to be make running the app locally easy in the three main environments.

The problem with it is that while we can say with reasonable amounts of certainty that there is one "dev" and one "test" environment, there isn't really one "prod" environment. From your 12-factor link:

> Sometimes apps batch config into named groups (often called “environments”) named after specific deploys, such as the development, test, and production environments in Rails. This method does not scale cleanly: as more deploys of the app are created, new environment names are necessary, such as staging or qa. As the project grows further, developers may add their own special environments like joes-staging, resulting in a combinatorial explosion of config which makes managing deploys of the app very brittle.

On you points:

> it will allows us to implement other specific config settings needed for staging deployment as the needs arise, rather than having to refactor more settings into env variables.

This is only easier if we never have another environment grouping.

> all non sensitive config is held in one place

It's held in at least two places at the moment (default and whatever environment file). 

> and couldn't find any reference of other benefits

Keeping config in _source-controlled_ files is bad because changes to config values shouldn't really require changes to code. The current solution is a halfway-house making concessions to the developer experience.

A better way, as described by the gist you linked, would be for the app accept a command line argument, pointing it to a config file which it ingests. But this is problematic when using heroku, since it has an ephemeral file system. Hence the env vars, which is heroku's [recommended way](https://devcenter.heroku.com/articles/config-vars) of managing config.

astroash:
> From your 12-factor link:
...
resulting in a combinatorial explosion of config which makes managing deploys of the app very brittle

This is an interesting point, but this article is arguing for exclusively using env vars. We already have an *explosion of configs* with our setup as we have both config files and env vars. I do not see anything mentioned here as a reason for separating out more of our config to env vars.

> > all non sensitive config is held in one place
>
> It's held in at least two places at the moment (default and whatever environment file).

For me this is the main reason I am dissatisfied with this solution. I consider our merged config files as one place for *non sensitive data*. This follows a pattern, it gives us a clear system for where config should be placed:

- 🔓 all non sensitive data is stored in environment config files that are merged with defaults
- 🔒 all sensitive data is stored in env vars

If I am to add a new piece of configuration to our server, it is clear where it should go

The suggested separation in this PR does not give a clear separation of where new configuration should go, or make it obvious where preexisting config is stored

> Keeping config in source-controlled files is bad because changes to config values shouldn't really require changes to code. 

This is valid...
> The current solution is a halfway-house making concessions to the developer experience.

but is not an argument for making our current half-way house less structured.

config/__tests__/index.test.ts
```diff
@@ -59,7 +59,7 @@ describe('Config', () => {
       router: { stripTrailingSlash: true },
       routes: {
         cors: {
-          origin: ['https://visitor.twine-together.com', 'https://admin.twine-together.com'],
+          origin: ['https://admin.twine-together.com', 'https://visitor.twine-together.com'],
```
astroash: Minor, just curious if there's a reason these two were switched around?

config/config.defaults.ts
```diff
@@ -40,6 +40,16 @@ export default {
       user: null,
       ssl: false,
     },
+    pool: {
+      min: 3,
```
astroash: Since we're setting this up, can we add a max?

Default is 10.

config/config.staging.ts
```diff
@@ -0,0 +1,22 @@
+import { Environment } from './types';
+
+export default {
+  env: Environment.STAGING,
+  web: {
+    host: process.env.HOST || '0.0.0.0',
+    port: process.env.PORT || 4002,
+    routes: {
+      cors: {
+        origin: ['https://twine-visitor-staging.herokuapp.com', 'https://admin.twine-together.com'],
```
astroash: I don't think we should add `'https://admin.twine-together.com'`. It won't work with our cookies, and there is no staging version of the app to point to this server. (Unless this is being added for testing)

config/config.defaults.ts
astroash: Typical defaults are 100. We can start with than and extend if needed.

PR #242, opened 2018-11-22
🐛 loosen password validation to support old volunteer app

### LINKED ISSUE #239: 🔒 password validation stronger than previous server
PHP app requires for passwords to be 6 characters

The new server requires passwords to be 8 characters long, include lower/upper case, numbers & symbols. This requirement is enforced on signup, login payload and clientside apps.

As the password requirements is stronger than previous synced users can not enter their "less secure" passwords.

astroash:
FIXES #239


eliascodes:
A thought: Can we also add "space" (` `) to the list of allowed special characters for the stronger password regex? It allows people to use sentences as passwords. `My old dog Gruff died in 1979` is a better password than `P@ssword1` and easier for people who are less comfortable with tech to remember.

PR #245, opened 2018-11-23
✨ add data_sync_log table

astroash:
**Changes:**
- create table with contraints
- add id to user_account_access_role so that it can referenced in data_sync_log table

relates TwinePlatform/data_sync#30


database/migrations/sql/20181123_16_add_data_sync_log_table.sql
```diff
@@ -0,0 +1,13 @@
+/*
+ * Data sync table
+ * - keep record of data synced from old dbs
+ */
+
+ CREATE TABLE data_sync_log (
+  data_sync_log_id  SERIAL NOT NULL UNIQUE,
+  foreign_key       INT NOT NULL,
+  table_name        CITEXT NOT NULL,
```
eliascodes: Any reason for `CITEXT` instead of `VARCHAR`?

astroash: Nope, can change

astroash: fixed in amend

PR #248, opened 2018-11-26
🐛 filter out "Extra Workspace" orgs

### LINKED ISSUE #237: Filter out "external workspace" orgs
These are dormant organisations, and should not be part of any API query results

eliascodes:
Relates #237 

This is intended as a temporary fix. See comment added to handler.


astroash:
@eliascodes can the named issue be changed from fixes -> related? This way the issue can be kept as a reminder of technical debt.

PR #260, opened 2018-11-28
GET /volunteer-logs returns username

astroash:
- allow names in volunteer logs for temp dashboard
  - update model VolunteerLog.get to return name
  - add userName to VolunteerLog type
  - fix tests
- allow names in logs for native app
  - support fields for volunteer logs from cbs
  - update /cb/me/vol-logs route to support fields
  - create util for converting request queries to model queries
  - update VolunteerLog.fromCb to support fields
  - update ModelQuery typings to support time queries

~~@eliascodes what does "bw" stand for as an arg? Only other small change if it doesnt fit VolLog methods being passed a ModelQuery~~ implemented

TwinePlatform/twine-temp-dash#3


eliascodes:
@astroash `bw` stands for `between` which can probably now be changed to `query` or `q` or something since we're now passing the full `ModelQuery` instead of a `DateModelQuery`.

Issue #247, opened 2018-11-26
Add support for multiple roles per user


eliascodes:
Many users have used the same e-mail to register as CB admins/volunteers/visitors because the old apps were running off of different DBs and had much weaker constraints.

The new data model does not allow duplicate e-mails, so multiple roles per user must be supported (otherwise we lose data).

eliascodes:
### API
* Credentials object needs changing, `roles` attribute becomes either an array or object of roles.
* `/users/me/roles` should be an array
* is_child_organisation
* is_child_user
* require_sibling_user
* Roles.oneFromUser
* Permissions possibly needs new convenience method
* Remove DB constraint on one role per user

### Volunteer app
* `/users/me/roles`

### Data sync
* Port multiple roles instead of overwriting.

astroash:
@eliascodes it's unclear from above how this is going to be implemented, but happy for you to do this as you see fit.

eliascodes:
It wasn't meant to be that detailed, just a summary of areas that need work.

eliascodes:
@astroash have just noticed that the list above (and the work I've done so far) doesn't include supporting new registrations of second roles (e.g. an existing volunteer signing up as a visitor). It only supports users who already have multiple roles due to the data-sync.

Do we want to support new registrations now or defer it?

astroash:
Defer for now.

PR #263, opened 2018-11-30
Multiple roles

eliascodes:
Related to #247 
Related to TwinePlatform/data-sync#33

### Changes
* Remove DB constraints on one-role-per-user. Now the only constraint on roles is that they cannot span multiple organisations.
* Update `Hapi.AuthCredential` typings to support multiple strategies
  * I think this was the source of the most number of lines changed, but the fewest actual functional changes. 
  * I did this because while working, I got to a point where the build worked but loads of tests failed, in a situation where IMO the compiler should have told me something was wrong. This was because the type constraints were too weak as a result of needing to support both `standard` and `external` strategies.
  * IMO we were [misusing the `credentials`](https://hapijs.com/api#-routeoptionsauthaccessentity) object before. We are now separately storing credentials for `standard` in `request.auth.credentials.user` and credentials for `external` in `request.auth.credentials.app`, which allows for stricter type constraints.
* Define and use `Credentials` interface instead of directly constructing and accessing `credentials` objects.
  * This is a layer of abstraction used to decouple the code that relies on the credentials from the code that defines what they look like. In future we'll only need to change a couple of functions instead of a few hundred lines.
  * We now use this in tests to define credentials passed to `server.inject`, which means the tests are more "integration-y" than before, but we cannot construct exotic auth situations which are not supported by the test data or the data model (both of which are IMO good things).
* `/user/me/roles` now returns array of roles
* Force visitor registration to fail if a user with _any_ role exists already with the registration email. (Volunteer registration already does this)
* Make `Turret` a `VISITOR` as well in the test data (and fix resulting test failures)


eliascodes:
Just FYI I'm user testing this locally at the moment.

eliascodes:
Nothing appears to be horribly broken yet. Will move onto another issue and come back to testing later.

src/api/v1/users/roles.ts
```diff
@@ -25,10 +25,14 @@ const routes: Hapi.ServerRoute[] = [
       } = request;
 
       const { user, organisation } = Credentials.fromRequest(request);
+      const roles = await Roles.fromUser(
```
astroash: Can roles be taken from credentials?

src/api/v1/users/register/visitors.ts
```diff
@@ -64,7 +65,7 @@ export default [
        * (Can possibly eventually be removed into pre-requisites)
        */
       // Check user doesn't already exist
-      if (await Visitors.exists(knex, { where: { email: payload.email } })) {
+      if (await Users.exists(knex, { where: { email: payload.email } })) {
```
astroash: Do we want to support visitors signing up who have other types of roles?

eliascodes: See https://github.com/TwinePlatform/twine-api/issues/247#issuecomment-443182884

Will add comment to the code

Issue #256, opened 2018-11-27
🚨 Setup error logging for server


astroash:
Moving into production, it would be useful to save detailed error logs.

eliascodes:
For now we can just log errors in the pre-response hook. It would also be useful to log user ID with every request to help with diagnostics and monitoring.

eliascodes:
So for every request, we want:
* Origin
* User ID, if possible
* Request payload
* Timestamp
* Endpoint hit
* Status code
* Response size (bytes)
* User Agent? (for browser stats)

Anything else?

eliascodes:
Closing for now, specific problems with this can be raised separately.

PR #271, opened 2018-12-06
[WIP] ✨ add initial version of custom logger

eliascodes:
related to #256

This is not a complete PR (lack of tests), but given the latest e-mail from Edward, I really want to get error logging in there ASAP, and I'll work on getting full coverage on this in parallel.

@astroash if you agree, I'll please remove the `WIP` mark and review/merge as normal 🙏


src/server.ts
```diff
@@ -6,7 +6,7 @@ import v1 from './api/v1';
 import setup from './setup';
 import routes from './routes';
 import { Config } from '../config/types';
-import logger from './logger';
+import Logging from './services/logging';
```
astroash: Curios why this has been named a verb?

src/services/logging/formatters.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const errorOrder: (keyof LogLine)[] = ['errorMessage', 'errorStack'];
+
+const extractCookie = (c: string) => {
+  const { [standard.cookie.name]: token } = Cookie.parse(c);
+  return token;
+};
+
+const decodeToken = (c: string): { userId?: number, organisationId?: number } => {
+  let decoded: { userId?: number, organisationId?: number };
```
astroash: Do we have a type for cookie? Also assuming that these values should not be optional.

src/server.ts
eliascodes: will rename

src/services/logger/formatters.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  return decoded;
+};
+
+const getIds = compose(decodeToken, extractCookie);
+
+const extractOrigin = (headers: Hapi.Util.Dictionary<string>) => {
+  return ['x-forwarded-for', 'origin', 'referer']
+  .reduce((acc, k) => acc ? acc : headers[k], null) || '';
```
astroash: Is this meant to be a find type function where you look for either 'x-forwarded-for', 'origin' or 'referer' headers in order?

`.find` may be more readable.

src/services/logger/formatters.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const errorOrder: (keyof LogLine)[] = ['errorMessage', 'errorStack'];
+
+const extractCookie = (c: string) => {
+  const { [standard.cookie.name]: token } = Cookie.parse(c);
+  return token;
+};
+
+const decodeToken = (c: string): { userId?: number, organisationId?: number } => {
+  let decoded: { userId?: number, organisationId?: number };
```
astroash: Do we have a type for cookie that we can use here instead?

Also is there a reason are these values optional?

src/services/logger/index.ts
```diff
@@ -11,26 +12,19 @@ const options = {
   reporters: {
     console: [
       {
-        module: 'good-squeeze',
-        name: 'Squeeze',
-        args: [{ log: '*', response: '*' }],
-      },
-      {
-        module: 'good-console',
+        module: GoodLogger,
```
astroash: Can we remove `good-squeeze` from our modules?

src/services/logging/formatters.ts
eliascodes: They are optional because if the token is invalid, `verfiy` will throw and we simply return an empty object (because we still want the logging to continue).

A little hacky in terms of the types but I just want logs now, I can tidy it up later.

src/services/logger/formatters.ts
eliascodes: `['a', 'b', 'c'].find(k => headers[k])` would return `a|b|c` instead of the value of those headers.
`Object.keys

src/services/logger/formatters.ts
eliascodes: Answered here: https://github.com/TwinePlatform/twine-api/pull/271#discussion_r239504990

src/services/logger/index.ts
eliascodes: Yes

src/services/logger/formatters.ts
eliascodes: Yes, can use that instead.

Second question answered here: https://github.com/TwinePlatform/twine-api/pull/271#discussion_r239504990

eliascodes:
Merging with permission from @astroash

eliascodes:
Merging with permission from @astroash

eliascodes:
Merging with permission from @astroash

PR #273, opened 2018-12-06
Merge STAGING to MASTER

eliascodes:
This is just pulling in the logging changes.

Going to self merge since @astroash is out of action and this is (IMO) high priority.


eliascodes:
OK this was way too premature, I've rolled back the deployment to the last one on discovering some issues.

🙈

PR #275, opened 2018-12-07
Logging

eliascodes:
Related to #256 

OK I fucked up the last one. This is more reliable and gives us the information we want.

### Changes
* Swap out `good` for `hapi-pino`
* Add tests on the serializers to hopefully rescue our coverage stats 😬


eliascodes:
@astroash checked on the review app and it seems to be working.

src/services/logger/serializers.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return decoded;
+};
+
+const getIds = compose(decodeToken, extractToken);
+
+const attachUserId = (req: { headers: Hapi.Util.Dictionary<string> }) => {
+  const { userId, organisationId } = getIds(req.headers);
+  return { ...req, userId, organisationId };
```
astroash: `userId` & `organisationId` are used in a number of payloads and responses, so it will be difficult to filter on this. I think we should make a unique string we can search on eg `auth_user:1`

Issue #259, opened 2018-11-27
add jti claim to JWT


eliascodes:
So we can revoke tokens if and when...

astroash:
@eliascodes did you mean jti here? This is a unique identifier for each jwt. From memory we were discussing having a version number attached to cookies so that they can be revoked on mass if our api changes.

eliascodes:
I think the wording of even the actual RFC is ambiguous, but from this [SO answer](https://stackoverflow.com/questions/43023218/jwt-jti-uniqueness), looks like you're right.

astroash:
Cool, just checking I had implemented it correctly

PR #278, opened 2018-12-07
🍪 add version number to jwt

astroash:
#259

@eliascodes this may or may not be what you had in mind


astroash:
@eliascodes in the end I don't think the typings needed updating. We want the create method to define the version number every time a cookie is created. It shouldn't be passed from previous payloads.

eliascodes:
Then it needs to be `TokenManager.create` that defines the version number, not the `SessionManager.create`.

PR #65, opened 2018-07-13
Linting

eliascodes:
~~**MERGE TwinePlatform/eslint-config-twine#10 FIRST**~~
~~**MERGE #64 FIRST**~~

Related to #51 
Related to #25 

Fixes a bunch of linting errors.

PR #283, opened 2018-12-12
Duplicate projects

eliascodes:
Related to TwinePlatform/twine-api#65

### Changes
* Prevent trying to set FK refs to columns that are soft deleted
* Prevent adding duplicate projects within a CB

### DIFF
--- a/src/models/volunteer_log.ts
+++ b/src/models/volunteer_log.ts
@@ -302,7 +302,7 @@ export const VolunteerLogs: VolunteerLogCollection = {
       ]);
 
       if (exists) {
-        throw new Error(`Project name ${project.name} is a duplicate`);
+        throw new Error(`Project name is a duplicate`);
       }
     }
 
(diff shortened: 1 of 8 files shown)


src/models/volunteer_log.ts
```diff
@@ -262,6 +262,16 @@ export const VolunteerLogs: VolunteerLogCollection = {
...
 
   async addProject (client, cb, name) {
+    const { rows: [{ exists }] } = await client.raw('SELECT EXISTS ?', [
+      client('volunteer_project')
+        .select()
+        .where({ organisation_id: cb.id, deleted_at: null, volunteer_project_name: name }),
+    ]);
+
+    if (exists) {
+      throw new Error('Cannot add duplicate project');
```
astroash: This error should be caught in the handler and return a 409

src/models/volunteer_log.ts
```diff
@@ -280,6 +290,22 @@ export const VolunteerLogs: VolunteerLogCollection = {
...
+          .select()
+          .where({
+            organisation_id: project.organisationId,
+            deleted_at: null,
+            volunteer_project_name: changeset.name,
+          }),
+      ]);
+
+      if (exists) {
+        throw new Error(`Project name ${project.name} is a duplicate`);
```
astroash: same as above

PR #290, opened 2018-12-29
✨ more filter query params for CB visitors & seed data

eliascodes:
Related to TwinePlatform/twine-visitor#536 and TwinePlatform/twine-visitor#562

### Changes
* Add email, post code and phone number to filter query params on `/community-businesses/me/visitors`
* Add seed data to demonstrate conflict scenarios for the "sign in with name" feature.

**Partnered PR**: TwinePlatform/twine-visitor#653


eliascodes:
@astroash Do you know if I need to do anything manually to get this deployed to a review app? Doesn't seem to have done it automatically even though the build and coverage checks passed.

database/seeds/test_data/12_test_visit_demo_data.seed.js
```diff
@@ -201,6 +201,242 @@ exports.seed = async (client) => {
...
+      /*
+       * Users intended as visitors to demonstrate various conflicts in the "Sign in with name" feature
+       */
+
+      // 01. name -> postCode -> GO  (Dupe name, unique postCode)
+      //
+      // 01 (foo,  SH3 4ND,  null,      null,        foo1@example.com),
+      //    (foo,  BR8 9EE,  null,      null,        foo2@example.com),
+      {
+        user_name: 'foo',
```
astroash: Can we get more usable names for this test data. I can imagine Edward using these to demo the app to new subscribers.

Something like 'visitor a' ect.

eliascodes: Happy to make the names easier to remember, but I suspect having super-fake values for post code/phone number might eventually come back to bite us if we add stricter validation on those fields.

astroash:
Can the demo data be moved to a migration rather than test_data? Currently it will not be added to the production db.

eliascodes:
> Can the demo data be moved to a migration rather than test_data?

I now think it was a mistake to use migrations as a way to ensure seed and/or demo data is inserted. It's convenient to begin with, but the conflation between structure and data will hurt us in the end IMO.

I think we should have much more control and flexibility over which demo data-sets are inserted where and when. We should have separate scripts to manage this.

astroash:
@eliascodes let's add it in here for now, then we can create an issue for restructuring migrations after this PR.

eliascodes:
I don't really want to, it'll just be more stuff to undo later on. Adding the data is simple enough, we can do it manually, I'll write something quickly. Nothing is relying on this data except for the ability to easily demo the functionality. IMO it's not blocking.

astroash:
@eliascodes yeah I am happy with that. I am just waiting to test the review app with the demo data.

Let me know when you've had a chance to do this.

eliascodes:
New visitors added to the staging DB now

Issue #289, opened 2018-12-19
Allow birthYear as an option field for users


astroash:
- [x] creating visitors should take birthYear as an optional field
- [x] visitor filters should take ~~'prefer not to say'~~ `null` as an age range 
- [ ] aggregated data should return data for ~~'prefer not to say'~~ `null` age range 

TwinePlatform/hq#40
TwinePlatform/twine-visitor#584

astroash:
Entering 'prefer not to say' as `null` in the database for birthYear will then require us to map all user queries to enter this value.

Currently user data is being pulled from 6+ model methods. As this grows I can easily see this being forgotten about and returning inaccurate data to clients. I am keen to change the birthYear column type from `int` to `text` in the database to allow us to enter 'prefer not to say'.

@eliascodes thoughts?

eliascodes:
Not massively keen on this as I think it conflates a presentational issue with a data representation issue.

`"Prefer not to say"` is just what we want the user to see instead of `NULL`. It's the clients who should be dealing with basically all presentational logic. A "null" value in `birthYear` should be interpreted as `prefer not to say`, where appropriate, by the client.

eliascodes:
Actually, my bad, another place this could be handled would be in the `User.serialise` method. We'd need to double check the handlers to make sure they're actually using this method before sending responses though.

I'm not opposed to the above because the `serialise` method is specifically meant for transforming models to being consumable by (all) clients.

> should setting the age value to null be the responsibility of the handler or the User.add model method. Currently for optional gender field this is handled by a joi default value.

Which handler are we talking about here? The visitor register handler doesn't set a default value, that's taken care of by the data model (as it should be).

> should our optional gender value be dealt with in the same way - eg saved in the db as null and displayed as 'prefer not to say' by the client

I think since gender/disability/ethnicity already work, I'm not that keen on changing them. 

<details>
<summary>TLDR</summary>
<p>
[...]

PR #292, opened 2018-12-31
✨ allow birthYear to be null

astroash:
- test User.add works with null value
- update types
- update Cb.getVisitLogAggregates to get null value age group
- add test data for visitor with null age
- update all tests due to new user
- escalate/deescalate routes return `null` for consistency

#289


astroash:
@eliascodes if possible I would like to merge this before #300 as I would like to use the newly created test user seeds for writing tests on that PR. Creating an extra user for an org throws loads of other tests off which has been fixed in this PR.

database/migrations/sql/20180604_06_create_users.sql
```diff
@@ -44,7 +44,7 @@ CREATE TABLE user_account (
   email                            CITEXT UNIQUE,
   phone_number                     VARCHAR(20),
   post_code                        CITEXT,
-  birth_year                       INT,
+  birth_year                       INT DEFAULT NULL,
```
eliascodes: We've gone live now, we cannot change any of the migrations any more. Any changes must be done with a new migration.

astroash: 😭

PR #288, opened 2018-12-19
🖌 add script for roles

astroash:

PR #304, opened 2019-01-05
Scripts

eliascodes:
Related to TwinePlatform/data-sync#39

### Changes
* Add script to re-migrate user logs (and check for consistency after-the-fact)
* Add script to edit user roles
* Add script to clone db from dump-file (optional download from heroku)
* Change (most) scripts to Typescript using `ts-node`

<hr>

~~Only making this PR to expose the visit logs re-migration script (also the script to clone a db from a pg dump file).~~

~~The only file of interest in relation to TwinePlatform/data-sync#39 is ` bin/remigrate_visit_logs.ts` (see 0da330a).~~

~~I've run this script off a clone of the latest backup of the API database and the existing visitor app database, and am left with 3740 visit logs, compared with only 880 before (original visitor app has 3887, which sounds like close enough given the test accounts we've ignored; we can chase down the difference if we need to).~~

~~I figured you could work off this branch to run the scripts you need, without it needing to be merged, but if you'd prefer to get it merged, let me know and I can do a little tidying up!~~

~~PS: Only 2024dbb -> 0da330a were added in relation to TwinePlatform/data-sync#39, everything else was done a while ago and is only vaguely related. Happy to do some re-basing if you want to get only a subset merged.~~


astroash:
@eliascodes why has my `bin/index.js` file been removed?

eliascodes:
TS scripts can be run directly with `ts-node` (or with `npm run exc`).

e.g `ts-node ./bin/create_user.ts` or `npm run exc -- ./bin/create_user.ts --name=...`

bin/edit_user_role.ts
```diff
@@ -0,0 +1,38 @@
+/*
```
astroash: Is this a script you have used on the production db?

bin/edit_user_role.ts
```diff
@@ -0,0 +1,38 @@
...
+ * script for editing an existing user
+ */
+import * as parse from 'minimist';
+import * as Knex from 'knex';
+import { Roles } from '../src/auth';
+import { getConfig } from '../config';
+
+process.on('unhandledRejection', (err) => { throw err; });
+
+const { userId, oid, add, rm } = parse(process.argv.slice(2));
```
astroash: Can you add comments for how this script should be used. Eg what args are passed to it.

bin/remigrate_visit_logs.ts
```diff
@@ -0,0 +1,186 @@
...
+ */
+import * as parse from 'minimist';
+import * as Knex from 'knex';
+import { Visitors, Organisations } from '../src/models';
+import { getConfig } from '../config';
+
+
+process.on('unhandledRejection', (err) => { throw err; });
+
+const { _: [sourceUrl], ignore, ignoreNullAct } = parse(process.argv.slice(2));
```
astroash: Does this work? You're using object destructuring on an array...

bin/remigrate_visit_logs.ts
```diff
@@ -0,0 +1,186 @@
...
+        console.log(
+          `${visitor.id}@${org.name} :: `,
+          `Source logs: ${sourceLogsPreSync.length} | `,
+          `Target logs: ${trgtLogsPostSync.length}`,
+          `Inserting ${payload.length} logs`
+        );
+
+        try {
+          console.log(`Visitor ${visitor.qrCode} has ${trgtLogsPostSync.length} new logs`);
+          await trx('visit_log').del().where('user_account_id', visitor.id);
```
astroash: What's going on here?

database/index.js
```diff
@@ -97,14 +97,14 @@ exports.migrate = {
 };
 
 exports.add = {
-  apiToken: async (name, access, { env = process.env.NODE_ENV, client: _client } = {}) => {
+  apiToken: async (name, access, token, { env = process.env.NODE_ENV, client: _client } = {}) => {
```
astroash: This can probably be removed due to the decoupling of frontline.

bin/edit_user_role.ts
eliascodes: No. It started off as a script to edit user details but became too complicated, so I just changed it to manage roles. It's similar to #288 but takes arguments.

eliascodes: Can also just delete though 🤷🏼‍♂️

bin/remigrate_visit_logs.ts
eliascodes: It works, the value of `_` is an array, so I'm using both array destructuring and object destructuring.

bin/remigrate_visit_logs.ts
eliascodes: Deleting all existing visit logs of the given visitor (re-added, along with any missing logs, on the next line).

bin/edit_user_role.ts
astroash: Up to you. Keep it if you think it’ll be useful in future.

eliascodes: Lets keep it for now then

bin/edit_user_role.ts
astroash: @eliascodes can this be commented?

PR #307, opened 2019-01-09
Fixes

### LINKED ISSUE #293: 🐛 neo-millenium bug
Custom joi schema for date is failing. `startedAt` constraint only works if the dates are in the same year

eliasmalik:
Fixes #293 

### Changes
* Fix bug in `rndPastDateThisMonth`
* Run `server_setup` script in test environment.

@astroash you can re-base #292 onto this and the tests should pass (also they should pass anyway because we're more than 2 days from the beginning of the month).


astroash:
I reran the tests on my PR and they passed. @eliascodes did you check this by changing your system date? (assuming that's where JS is getting time from).

Happy for this to be merged now, but wondering whether it's worth setting a reminder for 1st Feb to double check this so it doesn't end up being a recurring test bug like that log file

eliascodes:
> did you check this by changing your system date?

No, I fixed this on locally on the 1st, so I was still in the range where the bug would cause the problem (intermittently).

> whether it's worth setting a reminder for 1st Feb

Up to you. I think it's fixed, but happy to double check.

PR #308, opened 2019-01-09
🔧 change max DB connections to 20

### LINKED ISSUE #303: 🗃 check max connections for role on postgres server
When writing to the live database the following message was displayed:

```bash
too many connections for role "[DB_ROLE]"
```

Can be avoided by using transactions, but the limit should not be so low. Approx 20 connections were made.

eliascodes:
fixes #303


astroash:
@eliascodes, happy for this to be merged, but I don't think it "fixes" anything. It just makes it explicit in our code how many connections we have 😅

eliascodes:
It prevents 500 errors due to `knex` trying to allocate more connections than it is allowed.

astroash:
Do you know how `pg`s pool responds if we request more than 20 connections? Does it give us another error, reject the query/promise, wait for an open connection, something else?

Apologies if this is a simple question, I can't seem to find the answer online.

eliascodes:
We'll never request more than 20. `knex` simply requests _a_ connection from the pool. If all the connections in the pool are being used, the pool creates a new one. If the pool is already at the connection limit, it simply waits until a connection is returned to the pool.

astroash:
It may be worth us also setting `connectionTimeoutMillis` so that there's a max waiting time for a new client. We shouldn't run into this problem if _everything is working as it should_, but it will avoid our app hanging if there's a problem with clients disconnecting & make it easier to diagnose what's going on. 

https://github.com/brianc/node-postgres/issues/ 805

(☝🏽 i put a space in the url so i stop linking our repo to random issues)

eliascodes:
Shall we say 1 second timeout?

astroash:
@eliascodes sounds good.

PR #316, opened 2019-01-15
🔁 15-01-19 Deployment

astroash:
Additional User flows to test:
- less logs
- test db max connections

Clientside tests:
- resend qr codes for users without emails
- reset password invalid token
- weak password error messages
- add user without birthyear
- add passwords with £ € symbols


astroash:
@eliascodes writing a script to add the emailed CB_ADMIN's would be a great way to test the max connections...

eliascodes:
Note: there is client side validation on the native app that prevents the use of `£` and `€` for the creation of new volunteers by admin users. 
https://github.com/TwinePlatform/twine-app/blob/b39c75261679c03b2db6124cea23faa593b2109f/www/templates/new-volunteer.html#L31

`£` and `$` are, however, permitted for self-registration of new volunteers.

astroash:
@eliascodes can you make an issue/PR to Twine-app to fix the issues concerning symbols

astroash:
Sweet looks like this PR is ready. Let's wait until TwinePlatform/twine-visitor#628 is ready for merge.

PR #317, opened 2019-01-16
🔧 make file resolution for ts-node match tsc

eliascodes:
Fixes `npm run exc` build issues.


package.json
```diff
@@ -48,7 +48,7 @@
     "watch": "tsc -w",
     "build": "tsc && npm run copy",
     "heroku-postbuild": "npm run build && npm run migrate:latest",
-    "exc": "ts-node"
```
astroash: Can we call this script `exec`?

astroash:
@eliascodes please review and merge

bin/edit_user_role.ts
```diff
@@ -17,9 +17,11 @@ import { getConfig } from '../config';
 
 process.on('unhandledRejection', (err) => { throw err; });
 
-const { userId, oid, add, rm } = parse(process.argv.slice(2));
+const { _: args } = parse(process.argv.slice(2));
+const [userId, orgId, add, rm] = args.map((x) => x.split('=')[1]);
```
eliascodes: Ah, I see that I've missed something from the docs.

Usage is actually as follows:
```
$ npm run exec FILEPATH -- [--userId=a] [--orgId=b] ...
```
Note the `--`, which causes `npm` to pass the arguments following it to the process it executes rather than trying to consume them itself.

Including the `--` means that the arguments are correctly parsed. I can update the comment block at the top of the file to reflect this (and also at the top of the other scripts).

astroash: Makes sense. The additional `--` does make this script setup more tedious. IMO it will be simpler to add values directly into scripts.

PR #318, opened 2019-01-18
💫 register/visitors update

### LINKED ISSUE #298: 👱🏻‍♂️ Supporting anonymous user accounts
Some organisations create anonymous users so that they can scan the QR codes when people attend events.

eg:
anon woman
anon man
anon child

Currently this feature is not supported as user accounts need to be associated with an email. As a short term fix emails could be entered as a string such as `anon_[number]_org_[orgId]`. This would mean that emails will not be able to be sent for these accounts but their QR codes will be accessible from the visitor admin dashboard.

TwinePlatform/twine-visitor#607
TwinePlatform/data-sync#39

astroash:
💫 register/visitors to allow anon user creation & number OR email to identify

- update route to take isAnonymous boolean
- refactor route to use Visitor.add(Anonymous)WithRole methods
- allow either phone number or email as identifying data
- add tests for new use cases

fixes #298 
TwinePlatform/twine-visitor#536

**Please squash and merge**

User Flows:
- _see partnered PR_


eliascodes:
Are we still writing user flows to test in PRs? I mean I can guess which ones are important here so I guess it's not really necessary.

PR #321, opened 2019-01-28
✨ add script to fetch papertrail logs

eliascodes:
So they can be parsed and manipulated locally instead of via the Papertrail interface (sometimes that's easier).

This is mostly adapted from a code snippet in Papertrail's own docs (see https://help.papertrailapp.com/kb/how-it-works/permanent-log-archives/#download-a-large-number-of-archives).

### DIFF
--- a/bin/fetch_papertrail_logs.sh
+++ b/bin/fetch_papertrail_logs.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+#
+# Use to download archived logs from papertrail between two dates
+#
+# Usage
+#  $ ./bin/fetch_papertrail_logs.sh [from] [to] [output_file]
+#
+# Arguments
+#  from - YYYY-MM-DD date string; lower date limit for log retrieval
+#
+#  to - YYYY-MM-DD date string; upper date limit for log retrieval
+#
+#  output_file - filepath in which to dump the contents of the logs
+#
+# Environment
+#  PAPERTRAIL_API_TOKEN - [required] Personal API access token for Papertrail API
+#
+# NOTES:
+# - Ensure the script has execute permissions, otherwise execute via `bash`.
+# - All *.tsv.gz files in the root directory are removed afterwards. Backup any files you wish to keep
+
+# Variables
+DATE_FROM=$1
+DATE_TO=$2
+FNAME=$3
+PWD=$(pwd)
+PATTERN="${PWD}/"*.tsv.gz
+
+# Initial checks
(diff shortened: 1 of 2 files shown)


bin/fetch_papertrail_logs.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+#
+# Use to download archived logs from papertrail between two dates
+#
+# Usage
+#  $ ./bin/fetch_papertrail_logs.sh [from] [to] [output_file]
+#
+# Arguments
+#  from - YYYY-MM-DD date string; lower date limit for log retrieval
```
astroash: @eliascodes does this default to beginning or end of day?

bin/strip_logs.ts
```diff
@@ -0,0 +1,50 @@
...
+ *  npm run exc ./bin/strip_logs.ts < cat FILENAME
+ *
+ * Accepts no arguments, input taken from stdin.
+ * Use pipe (|) or redirection (<) to provide input.
+ */
+import { evolve, compose, omit, pick } from 'ramda';
+import { Stream } from 'stream';
+
+
+const readStreamToMem = (s: Stream): Promise<string> =>
```
astroash: Does this use `Stream` because we are piping the data file?

astroash:
@eliascodes you mention this is for maniplating logs? Would this manipulation be manually done in `strip_logs.ts`?

bin/fetch_papertrail_logs.sh
eliascodes: AFAIK beginning of day

bin/strip_logs.ts
eliascodes: Yes, `stdin` is a `ReadStream`

eliascodes:
> Would this manipulation be manually done in strip_logs.ts?

`strip_logs.ts` is only really meant to get rid of useless fields from the JSON logs. The manipulation we want to do would be based on what we're trying to find out.

Eventually it would be cool to do something like:
```
$ bash ./bin/fetch_papertrail_logs.sh YYYY-MM-DD YYYY-MM-DD | nr exec ./bin/strip_logs.ts | nr exec ./bin/do_some_custom_processing.ts > output.log
```
But I couldn't quite get the piping out of the first script to work for some reason.

astroash:
@eliascodes what do you want to do with this PR?

eliascodes:
We can retrieve this if we really need it.

PR #322, opened 2019-01-29
✨ add cb/register route

astroash:
**Changes:**
- add route that creates org, cb, admin, user & sends invite email
- add cbAdmin.addWithRole model method
- add route and model tests

relates TwinePlatform/hq#53

Partnered PR TwinePlatform/twine-temp-dash#13 👻 

Testing flows:
- see partnered PR


src/api/v1/community_businesses/register.ts
```diff
@@ -0,0 +1,97 @@
...
+      description: 'Register a new community businesses and invite admin user',
+      auth: {
+        strategy: 'standard',
+        access: {
+          scope: ['organisations_details-child:write'],
+        },
+      },
+      validate: {
+        payload: {
+          ...omit(['name'], cbPayload),
```
eliascodes: You can just use `renameKeys` on this, no? But also, is this even necessary? There's no naming conflict, and (IMO) it's fairly clear that `name` would mean the name of the CB 🤷🏼‍♂️

src/api/v1/community_businesses/schema.ts
```diff
@@ -39,4 +39,16 @@ export const visitActivitiesPutPayload = {
...
+  name: Joi.string().min(1),
+  region: Joi.string().min(1),
+  sector: Joi.string().min(1),
+  logoUrl: Joi.string().uri(),
+  address1: Joi.string().min(1),
+  address2: Joi.string().min(1),
+  townCity: Joi.string().min(1),
+  postCode: Joi.string().min(6).max(10),
+  turnoverBand: Joi.string().min(5).max(11),
+  _360GivingId: Joi.string().min(5).max(11), };
```
eliascodes: Where does the upper limit of 11 come from? From what I can see [there's no guarantee](http://standard.threesixtygiving.org/en/latest/identifiers/#) that IDs won't be longer than 11 characters. IMO let's make it the same as the DB constraint (255), or just remove the upper limit.

src/api/v1/community_businesses/__tests__/register.test.integration.ts
```diff
@@ -0,0 +1,118 @@
...
+  let admin: User;
+  let adminCreds: Hapi.AuthCredentials;
+  let organisation: Organisation;
+  const config = getConfig(process.env.NODE_ENV);
+
+  beforeAll(async () => {
+    server = await init(config);
+    knex = server.app.knex;
+
+    server.app.EmailService = {
```
eliascodes: Fancy setting up a simple mock of the `EmailService`? I think it would speed up the runtime of some of the tests.

Something like:
```ts
import MockEmailService from '../../tests/mocks/EmailService';

describe('', () => {
  let server: Hapi.Server;

  beforeAll(async () => {
    server = await init(config);
    MockEmailService.mock(server);
  });

  afterAll(async () => {
    MockEmailService.restore(server);
  });
})
```

src/api/v1/community_businesses/register.ts
astroash: `cb/register` payload also take a value for adminName. The naming convention here is to explicitly state which "name" is being referenced.

Can make this rename keys, had issues with ramda in the past so by default try to stay away from it in ts files 🤷🏽‍♀️

src/api/v1/community_businesses/schema.ts
astroash: addressed in #327

src/api/v1/community_businesses/__tests__/register.test.integration.ts
astroash: Are you suggesting moving L27-L40 to another file? Or to create a jest mock function?

I wonder if we could initalise the server in the global config n do something like `jest.mock('server.app.EmailService')`?

Unsure how this will speed up tests though, lemme know what you think n I can look into it further.

PR #329, opened 2019-01-30
🚀 DEPLOYMENT 30-01-19

astroash:
Changes:
- creates `cb/register`
- fix bug where you couldn't update turnover band to +70mil

Cannot be tested on hosting as there is no staging for `twine-temp-dash`

Partnered PR: TwinePlatform/twine-temp-dash#13 👻

User flow on `twine-temp-dash`
- can create a cb & admin user
- can update a cbs turnover band to +70mil

PR #330, opened 2019-01-30
🐛 return error for duplicate email

astroash:
Changes:

- return error when registering a new cb if email already exists
- add tests

Bug found as part of #329


src/api/v1/community_businesses/register.ts
```diff
@@ -60,10 +60,16 @@ export default [
         return Boom.forbidden('Insufficient permissions to create organisations');
       }
 
+      // Initial checks
       if (await CommunityBusinesses.exists(knex, { where: { _360GivingId } })) {
         return Boom.badRequest('360 Giving Id already exists');
       }
 
+      if (await Users.exists(knex, { where: { email: adminEmail } })) {
+        return Boom.badRequest('User already exists with this email');
```
eliascodes: Should be a 409 (conflict)

Issue #335, opened 2019-02-06
🔥 remove access_mode


astroash:
`access_mode` was added as a compromise for allowing cross site cookies for cb_admin and restricting access to the visitor app. This has caused issues, and it potentially no longer relevant as the `super_admin_dashboard` has been benched.

Moving forward we should remove this concept, and replace with additional roles added to users if needed in future.

TwinePlatform/twine-visitor#634

PR #336, opened 2019-02-06
🔥 remove access_mode

### LINKED ISSUE #170: Authentication routes

astroash:
- remove current data in db
- update db constraint 
- remove column in db
- remove inferprivilate from origin of requests
- remove privilege from cookies and schema
- emove escalate and deescalate

#335
fixes #170

**Partnered PR:** https://github.com/TwinePlatform/twine-visitor/pull/676
**Testing flows:** this is a permissions restructure EVERYTHING NEEDS TESTING


src/api/v1/__tests__/auth_flow.test.integration.ts
```diff
@@ -43,75 +43,6 @@ describe('Auth flow', () => {
 
       expect(resAccess.statusCode).toBe(200);
     });
-
-    test('CB login from visitor :: escalate session :: de-escalate session', async () => {
```
eliascodes: The `describe` block here is still called `Escalation`. Should probs be changed.

Issue #338, opened 2019-02-08
🖋 script for demo data


astroash:
Write script to add demo data outlined in #290 to production database

eliascodes:
I want to use this as an opportunity to separate seed data from test/demo data. The former is just any data that is necessary for the normal operation of the app (e.g. list of supported disabilities). Test/demo data is any additional data that is required only for testing or demo purposes.

They should be defined separately. I was imagining something like:
```
database
|- migrations
|- seeds
|- static
|- test_data
|  |- demo_organisation_duplicate_visitors
|  |- demo_organisation_visitor_data
|  |- demo_organisation_volunteer_data
|  |- test_organisations
|  |- test_users
|  |- test_user_roles
|  |- ...
|- util
```
Here I'm using `test_*` or `demo_*` prefixes to distinguish between data used for demo and data used for tests.

We'd then use a command line tool to insert/remove various data sets:
```
$ db data insert [DATA_SET/FILENAME]
$ db data reset [DATA_SET/FILENAME]
```

Thoughts?

astroash:
@eliascodes how do you feel about implementing our own version of migrations for demo data? This would be a nice way to keep track of what demo data has been inserted in to staging/production dbs and make it easier to recreate locally.

eliascodes:
Yeah that's a good idea!

astroash:
@eliascodes can you update this issue with your findings

eliascodes:
Steps that I wanted to take:
- [ ] Change everything to typescript
- [ ] Separate directories for test data and demo data
- [ ] Demo data can be attached to any specified CB

I am now less convinced of the usefulness of a data migrations table. Demo data will only be added to junk accounts, which could be identified using #361 

What is the benefit of being able to say "we've added demo set A to CB X"?

astroash:
Migration tables would be beneficial if we end up adding more demo data to demo accounts in the future to support features on new apps.

eliascodes:
> Migration tables would be beneficial if we end up adding more demo data to demo accounts in the future to support features on new apps

But how exactly?

astroash:
On reflection, I think you are right.

eliascodes:
Tasks outlined in [comment](https://github.com/TwinePlatform/twine-api/issues/338#issuecomment-469301895) should be covered by their own issues.

PR #341, opened 2019-02-13
✨ add script to insert demo/test data

eliascodes:
Related to #338 

This is part of moving towards a full TS stack (and retiring JS files/scripts).

This is implementing the beginning of the structure I sketched out in my comment: https://github.com/TwinePlatform/twine-api/issues/338#issuecomment-462345624

* Added brief docs explaining structure of test data
* Re-organise how seed and test data is stored and inserted into the DB
* Upgrade `jest` so that `global_setup` can be converted to TS
* Add simple CLI entry point to insert test/demo data (to be improved in future)

I had much more significant changes in mind, but figured I shouldn't run before I can walk.


eliascodes:
User testing:
* [ ] Reset local (dev) DB
   1. Run `npm run db:reset` -> Should have no data in DB (except gender, disability, etc)
   2. Run `npm exec bin/insert_data.ts demo` -> All demo data should be inserted
* [ ] Reset remote (staging?) DB (same steps as above, except setting `NODE_ENV`)
* [ ] Run tests

astroash:
**Manual testing feedback:**
- if we are using `exec` in this way it would be nice to get more visual feedback of what is happening eg rows inserted, complete ect. Currently the only feedback is:
```
> twine-api@1.0.0 exec <directory>/twine/twine-api
> ts-node --files "bin/insert_data.ts" "demo"
```
- migrations suggestion outlined in https://github.com/TwinePlatform/twine-api/issues/338#issuecomment-464654553

eliascodes:
> migrations suggestion outlined in [#338 (comment)](https://github.com/TwinePlatform/twine-api/issues/338#issuecomment-464654553)

I like this idea but I think it's a bigger piece of work.

astroash:
> I like this idea but I think it's a bigger piece of work

Are we in any rush to get this merged?

eliascodes:
Well, actually having the seed data on master/staging might be nice. Beyond that, I guess not. But also depends on how you feel about reviewing enormous PRs 🙂

astroash:
> But also depends on how you feel about reviewing enormous PRs 🙂

😅 I can make an exception.

I feel like this PR is v close to having a super sweet db setup. As all the changes are to scripts that we have to run manually I see little benefit of it being merged to staging/master. I am happy to wait for this to be amended to run as migrations.

That said, if you think the migration work is less of a priority than other things on our backlog then we can do it in stages (eg merge this now then get to migrations later). First set of demo data has already been added to production so we will have to fake the migrations table for those entries anyway.

astroash:
@eliascodes is this ready for review?

eliascodes:
I decided to remove the extra work because it's better to do it bit by bit to prevent me from losing my mind 🙃.

It's now at the same state as it was when you originally left your review, (except with some logging added).

bin/insert_data.ts
```diff
@@ -0,0 +1,35 @@
+/*
+ * DB CLI
+ */
```
astroash: can you add instructions on how this works

astroash:
@eliascodes please update the PR body and squash-merge 🙂

PR #342, opened 2019-02-14
🐛 fix validation bug preventing visit activity deletion

eliascodes:
The only valid payload was `null`, but actually, we don't use the payload at all, so can just remove validation.

Additionally, we were for some reason validating the query string, which only makes sense for `GET` routes.

Fixes TwinePlatform/twine-visitor#685


astroash:
@eliascodes can you add any manual testing needed for this PR

eliascodes:
User Testing:
- [ ] Delete activity

PR #344, opened 2019-02-16
🗃 replace update trigger function to ensure casing updates work

### LINKED ISSUE #333: Casing updates on CITEXT fields don't work
Discovered in TwinePlatform/twine-temp-dash#12

Reproducing my comment here:
> I think this is because we're using the `CITEXT` data type. It looks like postgres is comparing the value we're trying to set to the value that's already there, and if they match, it doesn't actually do the update.
>
> So if we were using a `TEXT` data type, it would compare, for example, `Twine` to `twine`, notice a difference, and then update the value. But using `CITEXT`, there's an implicit application of `LOWER()`, so it is actually comparing `LOWER('Twine')` to `LOWER('twine')`, which is the same.
>
> Perhaps unhelpfully, in both cases, the query returns `UPDATE 1`, telling the client that a row has been updated.
>
> I came to this conclusion by trying to update org names: 
>```sql
>UPDATE organisation SET organisation_name='Extra workspace 150' WHERE organisation_id=X;
>```
>This returns `UPDATE 1` but the name remains `Extra Workspace 150`. But if we change anything else, then the update "works":
>```sql
>UPDATE organisation SET organisation_name='Extra workspace 151' WHERE organisation_id=X;
>```
>Now the name is `Extra workspace 151`.

eliascodes:
Fixes #333

**Changes**
- replace update trigger function to ensure casing updates work
- This adds tests on casing changes to various API endpoints
- updates `PUT /community-businesses/me/visit-activities` to accept `name` in the payload, making it editable now (though this is not exposed by any front-end features).


eliascodes:
Manual testing:
- [ ] Run `npm run migrate:latest` on local DB
- [ ] Run version of twine-temp-dash against local DB, try to update casing on Org name

astroash:
@eliascodes is it worth adding tests for this trigger change? It would be useful to see that this change has the desired effects on various data type columns

PR #349, opened 2019-02-22
✨ Server Changes for Temp Accounts

astroash:
**Changes:**
- add temp values for region and sector
- create the following routes:
  - `POST /community-businesses/register/temporary`
  - `GET /community-businesses/{id}/cb-admins`
  - `DELETE /community-businesses/temporary/:id`
  - `GET /community-businesses/temporary`
  - `POST /community-businesses/register/temporary`
- add model method for `Cb.addTemporaryWithRole`
- add is_temp columns to org & user table
- add isTemp to CB type

**NOTE:** I have added a flow test for the delete method, tho after that password bug we had pre release, this level of hard deleting makes me nervous. Can you think of a simple way we can check if there have been any other db change pre > post test?

**Partnered PR:** https://github.com/TwinePlatform/twine-temp-dash/pull/15

fixes #349


database/migrations/sql/20190220_19_add_temp_region_sector.sql
```diff
@@ -0,0 +1,6 @@
+/*
+ * Migration template
+ */
+
+INSERT INTO community_business_sector (sector_name) VALUES ('TEMPORARY DATA');
```
eliascodes: Why is this necessary?

src/api/v1/community_businesses/temporary/delete.ts
```diff
@@ -0,0 +1,54 @@
...
+      if (!isTempAccount) {
+        return Boom.forbidden('Not a temporary organisation');
+      }
+
+      await knex.transaction(async (trx) => {
+        // delete all users at this organisation
+        await trx.raw('DELETE FROM user_account '
+        + 'USING user_account_access_role '
+        + 'WHERE user_account.user_account_id = user_account_access_role.user_account_id '
+        + 'AND user_account_access_role.organisation_id = ?', [communityBusiness.id]);
```
eliascodes: Does this need a `CASCADE`?

src/api/v1/community_businesses/temporary/delete.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        await trx('organisation')
```
eliascodes: Similarly, does this need a `CASCADE`?

src/models/cb_admin.ts
```diff
@@ -69,6 +70,26 @@ export const CbAdmins: CbAdminCollection = {
     });
   },
 
+  async addTemporaryWithRole (client, cb) {
+    const [latestTempCbAdmin] = await client('user_account')
+      .select('*')
```
eliascodes: select `email`? that's the only thing we need here

src/models/cb_admin.ts
```diff
@@ -69,6 +70,26 @@ export const CbAdmins: CbAdminCollection = {
...
+      .where('email', 'like', 'welcome-%@twine-together.com')
+      .orderBy('created_at', 'desc')
```
eliascodes: what happens if we sort by `email`? Wouldn't the biggest number then be first/last?

src/api/v1/community_businesses/temporary/get.ts
```diff
@@ -0,0 +1,41 @@
...
+      ],
+      response: { schema: response },
+    },
+    handler: async (request: Hapi.Request, h: Hapi.ResponseToolkit) => {
+
+      const {
+        server: { app: { knex } },
+      } = request;
+
+      // TODO: stick in model method
```
eliascodes: Let's actually do this?

database/migrations/sql/20190220_19_add_temp_region_sector.sql
astroash: `sector_id` is a required foreign key for `community_business`. I wanted to attach it to its own row to avoid complications with aggregated statistics in the future

Happy for suggestions on a better name.

src/api/v1/community_businesses/temporary/delete.ts
astroash: How do you mean? I was under the impression `cascade` is covered by the table constraints not something implemented in queries.

database/migrations/sql/20190220_19_add_temp_region_sector.sql
eliascodes: It's better to represent things directly than indirectly. That means things can be declared rather than inferred.

If temp/dummy organisations are a long term feature, then we want them to be segregated by default from real organisations. Ideally that would mean the temp/dummy nature of an organisation should be a first-class property of the organisation model.

At the moment we're inferring that an org is temp by the name/email. This is a pragmatic step to avoid changing the data model, but I don't think that should be a long-term thing.

We should go through the CB model methods and make sure "normal" methods cannot return "temp" orgs, and "temp" methods cannot return "normal" orgs (and similarly for cb admin methods).

src/api/v1/community_businesses/temporary/get.ts
astroash: `CommunityBusinesses.getTemporary`?

Also do we want a separate type for temporary CB? Or happy for them to squeeze into a partial cb?

src/models/cb_admin.ts
astroash: Numbers would be seen as individual characters, so will only work up to 9.

src/api/v1/community_businesses/temporary/get.ts
eliascodes: `getTemporary` sounds fine. My initial reaction is that they don't need a separate type.

src/models/cb_admin.ts
astroash: @eliascodes happy for me to resolve this?

database/migrations/sql/20190220_19_add_temp_region_sector.sql
astroash: Is this in addition to having a temp value for sectors/regions? I do not think we need to collect this info for creating a temp account and I do not want to just throw them under an arbitrary value.

src/api/v1/community_businesses/temporary/get.ts
astroash: So, technically, the return value does not match an array of our `CommunityBusiness` type. I thought this was okay because all the values I am not returning here are ones we are leaving empty eg address1, postcode ect

I feel like it's inaccurate in this instance to say that the model method will return `[] Partial<Cb>`, I could pick the values I know it'll return but it's very specific to this use case.

🤷🏽‍♀️ any thoughts? I was keen to leave this in the handler until this feature becomes a bit better understood (and whether we add anything similar to help with onboarding/analytics)

database/migrations/sql/20190220_19_add_temp_region_sector.sql
eliascodes: > Is this in addition to having a temp value for sectors/regions?

It's instead of.

> I do not want to just throw them under an arbitrary value.

Why not? They only reason regions/sectors are needed is in order to be able to calculate statistics across and within these categories. If we can already segregate out the "fake" orgs, then the quality of the statistics is safe.

astroash: > Why not?

I do not feel that we should be storing data inaccurately in the db if it can be avoided. What do we loose by having temp values in region/sector?

src/api/v1/community_businesses/temporary/get.ts
eliascodes: I'm thinking that I'd like the models to be a place where we keep all the "questions" we are asking of the data. If they're all (roughly) in one place then it's easier to see patterns as and when they arise.

I'm currently not so bothered if the type signature of a few of the methods are super-specific.

database/migrations/sql/20190220_19_add_temp_region_sector.sql
eliascodes: Every aspect of these temp organisations is inaccurate

astroash: None of the data added is fake though.

astroash: Still keen to know:

> What do we loose by having temp values in region/sector?

eliascodes: > None of the data added is fake though.

What do you mean by "fake"? It's all intended for eventual deletion, none of the data associated to those temp accounts matters at all.

> What do we loose by having temp values in region/sector?

We _gain_ cruft, indirection, and workarounds.

I don't want to hold this up too much. I don't like this idea, but if you really want to leave them in, do.

But I would like this to be done:

> We should go through the CB model methods and make sure "normal" methods cannot return "temp" orgs, and "temp" methods cannot return "normal" orgs (and similarly for cb admin methods).

It makes sense for the definition of "temporary" account to be abstracted out somewhere so it can be easily changed in future.

I will create an issue about adding in a field to identify temp accounts/orgs directly (and relatedly, discussing what will happen to the `subscription` and `funding_body` tables), and then updating the temp CB model methods.

astroash: > What do you mean by "fake"?

Here's an example of all org/cb data we hold for a temp cb created as part of this PR. Any data that is not relevant for a temp account is set as `null`. I cannot set `region_id` & `sector_id` as `null` because of db constraints, so they have their own related row `TEMPORARY DATA` inserted into each table. 

```js
organisation_id: 14,
    organisation_name: 'TEMPORARY ACCOUNT: bla',
    _360_giving_id: null,
    created_at: 2019-02-22T17:24:20.710Z,
    modified_at: null,
    deleted_at: null,
    community_business_region_id: 10,
    community_business_sector_id: 15,
    address_1: null,
    address_2: null,
    town_city: null,
    post_code: null,
    coordinates: null,
    logo_url: null,
    turnover_band: null }
```

> But I would like this to be done:

In this PR?

eliascodes: > In this PR?

I was thinking yes, but actually maybe it's not that important yet. Can you think of any places where it might cause an issue for temporary and regular orgs to be treated exactly the same?

* Login - fine to treat them the same
* `GET /community-businesses/{id?}` is only for use by `TWINE_ADMINS` so far, so that's not a big deal
* `<ANY> community-businesses/me/{param?}` - fine already

Anything else?

eliascodes: OK, then I'll mark this resolved once those issues are created.

astroash:
TO DO:
- remove 🚐 rename db migration commit 😕

astroash:
@eliascodes heroku deployment is failing because of migrations, but this PR can only be tested locally as it is connected to `twine-temp-dash` #356

database/migrations/sql/20190220_19_add_temp_region_sector.sql
eliascodes: Created #361 #362 #363

src/api/v1/community_businesses/temporary/register.ts
```diff
@@ -0,0 +1,61 @@
...
+      } = request;
+
+      if (!isChild) {
+        return Boom.forbidden('Insufficient permissions to create organisations');
+      }
+
+      // Create accounts
+      const { admin, cb } = await knex.transaction(async (trx) => {
+        const cb = await CommunityBusinesses.add(trx, {
+          name: `TEMPORARY ACCOUNT: ${orgName}`,
```
eliascodes: I've just created two accounts with the same name by accidentally double clicking the button. I also created an account with the same name as an existing org.

So I think we should check for existence of orgs with the name `orgName` and `TEMPORARY ACCOUNT: ${orgName}`

src/models/community_business.ts
```diff
@@ -187,6 +187,18 @@ export const CommunityBusinesses: CommunityBusinessCollection = {
...
+  async getTemporary (client) {
+    return client('organisation')
+    .select({
+      id: 'organisation.organisation_id',
+      name: 'organisation.organisation_name',
+      createdAt: 'organisation.created_at',
+      modifiedAt: 'organisation.modified_at',
+      deletedAt: 'organisation.deleted_at',
+    })
+    .where('organisation.organisation_name', 'like', 'TEMPORARY ACCOUNT: %');
```
eliascodes: As soon as a temp account is created, I can log in as it into the visitor app, rename the organisation to anything that doesn't include `TEMPORARY ACCOUNT`, and suddenly it's a fully fledged organisation.

This is why I wanted things to be [represented directly](https://github.com/TwinePlatform/twine-api/pull/349#discussion_r260258368), because encoding things indirectly like this leads to unforseen issues.

We'll need to address #361 in this PR.

src/api/v1/community_businesses/temporary/register.ts
astroash: Why can't a CB have two demo accounts?

src/models/community_business.ts
astroash: 😬 I had not considered this!

src/api/v1/community_businesses/temporary/register.ts
astroash: @eliascodes I do not think it is necessary to restrict cbs to only one temp account. Let me know if you can think of any unintended side effects of this.

eliascodes:
IMO the long term fix for this is to have a first-class notion of "account type".

astroash:
> IMO the long term fix for this is to have a first-class notion of "account type".
 
can you put your thoughts in an issue

eliascodes:
The issue tag in the PR description is wrong: it references itself. It should reference TwinePlatform/hq#60.

Issue #348, opened 2019-02-21
👣 tracking user activity


astroash:
We are currently tracking login events. However as we have cookies that last for one month this can be deceptive as 1 login event can last for this period. I wonder whether it is worth us tracking "days active", so we create a log every day a user is active on the platform. 

@eliascodes let me know what you think of the technicalities of this before I open the discussion to the rest of the team.

eliascodes:
We could hook into the `onPreHandler` event to do this without polluting every route.

One row for every day the user is "active" is probably the least aggregated quantity we could record for now, though maybe worth thinking about whether "day" is definitely the minimum timeframe of interest (IMO it is, but...).

astroash:
Sweet, sounds like that that's a simple way to implement. I'll add this to our baseline agenda later. I'd also be keen to display this info on my current temp-dashboard changes for https://github.com/TwinePlatform/hq/issues/60

astroash:
confirmed by @Shanozd that this would be a useful data point to collect. Placing a high priority as this is a minimal change, and it would be good to start collecting asap

PR #352, opened 2019-02-27
✨ record user usage by day - addActiveDayEvent

astroash:
- create new db table
- add model method to add active day event
- add method to api hook
- add tests

#348 

**Manual testing**
- covered by tests :)


src/api/v1/hooks/on_pre_response.ts
```diff
@@ -14,11 +14,20 @@ import * as Hapi from 'hapi';
...
+import { Users } from '../../../models';
+import { StandardCredentials } from '../../../auth/strategies/standard';
 
 
 export default async (request: Hapi.Request, h: Hapi.ResponseToolkit) => {
   const env = request.server.app.config.env;
 
+  try {
+    const { user } = StandardCredentials.fromRequest(request);
+    await Users.addActiveDayEvent(request.server.app.knex, user, request.headers.origin);
```
eliascodes: We don't even need to `await` this. We just want to fire and forget really.

database/migrations/sql/20190226_21_add-active-days.sql
```diff
@@ -0,0 +1,15 @@
+/*
+ * Add active-
```
eliascodes: incomplete?

src/models/user.ts
```diff
@@ -319,4 +319,23 @@ export const Users: UserCollection = {
 
     return null;
   },
+
+  async addActiveDayEvent (client, user, origin) {
+    const [exists] = await client('user_account_active_day')
+      .where({
+        user_account_id: user.id,
+         origin,
```
eliascodes: how come this inconsistent indentation isn't caught by the linter?

astroash: cos palantir...

eliascodes:
Can you take a look at the deployment issue? Is it just migrations?

eliascodes:
Can you explain why you had to put the `await` back in?

astroash:
_my understand which may or may not be right_:

Knex uses bluebird which is lazy promises that only do the thing when you call then/await. 

Also breaks all the tests if you don’t.

astroash:
To do:

- check migration table names aren't wack
- add .then().catch() to lazy promise

astroash:
@eliascodes really struggling to test the hook properly with mocks, so hacked it 💩. Let me know if you have a better solution.

eliascodes:
What about:
```ts
await new Promise((resolve) => setTimeout(resolve, 100));
```
At least that's more explicit about whats happening.

database/migrations/sql/20190226_add-active-days.sql
```diff
@@ -0,0 +1,15 @@
+/*
+ * Add active days table for recording user interaction
```
eliascodes: just noticed that the migration files use dashes instead of underscores, let's use underscores to be consistent.

src/models/user.ts
```diff
@@ -319,4 +319,23 @@ export const Users: UserCollection = {
...
     return null;
   },
+
+  async addActiveDayEvent (client, user, origin) {
+    const [exists] = await client('user_account_active_day')
+      .where({
+        user_account_id: user.id,
+        origin,
+      })
+      .andWhereRaw('created_at >= CURRENT_DATE - INTERVAL \'1 day\'');
```
eliascodes: This means that the following could happen:
1. User request at 1pm day 0 => active event stored
2. User request at 3am day 1 => no active event stored

Is that what we want? The requests happened on two different days, shouldn't they both be stored?

eliascodes:
The test cases don't reflect the new condition.

src/models/__tests__/user.test.integration.ts
```diff
@@ -363,5 +363,89 @@ describe('User Model', () => {
...
     });
   });
+  describe('addActiveDayEvent', () => {
+    test('::SUCCESS - adds an event for a user', async () => {
+      const user = await Users.getOne(trx, { where: { id: 1 } });
+
+      // check no event exists
+      const eventCheck = await trx('user_account_active_day')
+      .where({ user_account_id: user.id })
+        .andWhereRaw('created_at >= CURRENT_DATE - INTERVAL \'1 day\'');
```
eliascodes: Missed a bunch of these conditions

PR #359, opened 2019-03-03
🐛 fix user scope

astroash:


eliascodes:
If this is only merged to master, doesn't beta need a similar PR, or to be re-based onto `master/HEAD`?

astroash:
@eliascodes yes. Any preference?

eliascodes:
No preference.

PR #370, opened 2019-03-19
✨ Support dashboard in pwd reset

eliascodes:
Related to TwinePlatform/volunteer-dashboard#7

### Changes
* Add new E-mail template
* Allow reset requests from dashboard app
* Remove snake-case from config parameter
* Type-check config fragments
* Add method to map App -> PWD_RESET template
* Lil bit of renaming

**Partnered PR**: TwinePlatform/volunteer-dashboard#6


config/config.defaults.ts
```diff
@@ -79,4 +80,9 @@ export default {
   qrcode: {
     secret: process.env.QRCODE_HMAC_SECRET,
   },
+  email: {
+    fromAddress: 'visitorapp@powertochange.org.uk',
```
astroash: 🎉 i was thinking of doing this in my pr

src/api/v1/community_businesses/volunteer_logs.ts
```diff
@@ -431,14 +431,7 @@ const routes: Hapi.ServerRoute[] = [
...
           where: { organisationId: communityBusiness.id },
-          whereBetween: since || until
-            ? {
-              startedAt: [
-                since || new Date(0),
-                until || new Date(),
-              ],
-            }
-            : undefined,
+          whereBetween: { startedAt: [since, until] },
```
astroash: how come the default `new Date` has been removed?

src/api/v1/users/password.ts
```diff
@@ -7,15 +7,15 @@ import {
   email as emailSchema,
   password as passwordSchema,
 } from './schema';
-import { EmailTemplate } from '../../../services/email/templates';
+import { Templates } from '../../../services/email';
 import { BoomWithValidation } from '../utils';
-import { AppEnum } from '../types';
+import { AppEnum } from '../../../types/internal';
```
astroash: Do you think it would be easier for us to setup a types folder in root? I feel like I often spend time wondering where the types are for various bits n pieces.

src/types/internal.ts
```diff
@@ -45,7 +47,15 @@ export type Day =
...
   | 'friday'
   | 'saturday'
-  | 'sunday'
+  | 'sunday';
+
+export enum AppEnum {
+  TWINE_API = 'TWINE_API',
+  VISITOR = 'VISITOR_APP',
+  VOLUNTEER = 'VOLUNTEER_APP',
+  DASHBOARD = 'DASHBOARD_APP',
```
astroash: I wonder whether it's worth calling this `DATA` rather than dashboard, then we're describing the subdomains. Or do you think thats too much coupling to hosting?

src/api/v1/community_businesses/volunteer_logs.ts
eliascodes: Because that branch was unreachable. `since` and `until` are both already `Date` objects (see line 426), so the `||` condition will never fall through.

src/types/internal.ts
eliascodes: Yeah, I think that's too much.

src/api/v1/users/password.ts
eliascodes: Yeah, TBH I wanted a way to make all our type definitions available globally (since they are ephemeral anyway, it doesn't matter), but could never figure out a way to do it.

One big root-level types file/folder might make sense. Not in this PR though :)

PR #371, opened 2019-03-19
✨ Add multiple-role confirmation flow

astroash:
Related to TwinePlatform/hq#57

**Changes:**
- update Roles.fromUser to be more specific
  - model method now gets all roles for any org from a user (not currently supported by apps but a useful feature)
  - old .fromUser renames to fromUserWithOrg as it requires an org id to work
- add new email template id
- /register/visitor sends confirmation email if user exists at cb
- /register/volunteer sends confirmation email if user exists at cb
  - only if registering as a volunteer
- /users/register/confirm takes token and adds new role
- add table for add role tokens

Partnered PR: https://github.com/TwinePlatform/twine-visitor/pull/704

Notes:
- sending a confirmation email returns a 409 status code, as there is a conflict but we perform an auxiliary action


src/api/v1/users/register/volunteers.ts
```diff
@@ -51,23 +53,77 @@ export default [
...
+
+        if (!orgMatchesCurrentRoles) {
+          throw Boom.conflict(
+            'User with this e-mail already registered at another Community Business');
+        }
+
+        const { token } = await Users.createConfirmAddRoleToken(knex, user);
+
+        try {
+          await EmailService.send({
```
eliascodes: I think there's room for an intermediate abstraction that sits between the generic parts of the `EmailService` and the handler (application-specific logic), in order to make the handler more declarative.

Something like:
```
EmailService.confirmAddRole(user, cb, token, role)
```

src/models/types.ts
```diff
@@ -304,8 +304,12 @@ type UsersBaseCollection = Collection<User> & {
 };
 
 export type UserCollection = UsersBaseCollection & {
+  createToken: (k: Knex, u: User, t: string) => Promise<SingleUseToken>
```
eliascodes: Maybe all these token methods should be moved onto a dedicated `Tokens` collection

src/api/v1/users/register/confirm.ts
```diff
@@ -0,0 +1,129 @@
...
+        // Because of a change in testing method, this is now unreachable
+        // (and was always functionally impossible), but is kept **just
+        // in case**
+        throw Boom.badData('No associated admin for this organisation');
+      }
+
+      // Checks
+      const currentRoles = await Roles.fromUser(knex, user);
+      // check role doesnt exist on user account
+      if (currentRoles.some((x) => x.role === role)) {
```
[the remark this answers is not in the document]
astroash: Roles.userHas does something slightly different:

```js
{
	role: RoleEnum
	userId: number,
    organisationId: number
} => boolean
```

The above grabs a list of all the roles a user has so that we can check if they are for the same organization.

src/api/v1/users/register/volunteers.ts
astroash: To be completed in separate refactor PR

src/models/types.ts
astroash: To be completed in separate refactor PR

src/auth/roles/index.ts
```diff
@@ -102,7 +108,23 @@ const Roles: RolesInterface = {
     return rows[0].exists;
   },
 
-  fromUser: async (client, { userId, organisationId }) => {
+  fromUser: async (client, user) => {
+    return await client
```
eliascodes: `return await` is redundant unless you're wrapping this all in a `try/catch`

src/models/user.ts
```diff
@@ -340,4 +357,10 @@ export const Users: UserCollection = {
 
     return;
   },
+
+  async hasCommunityBusiness (client, user, cb) {
```
eliascodes: `has` is an ownership indicator. `CommunityBusinesses.hasUser` would make sense, but if the method is staying here then `Users.isMemberOf` is better.

